PR #25, opened 2015-03-17
Geant sim

zaborowska:
HepMC -> G4event conversion (based on G4
examples/extended/eventgenerator/HepMC/HepMCEx01/src/HepMCG4Interface.cc )
and simple geant simulation (detector temporarily from G4 examples/basic/B1/)


hegner:
Hi Anna, 

thanks! I believe the HepMC to Geant Converter should go under Simulation/ as well...

  Benedikt

DataObjects/DataObjects/GeantEventEntry.h
```diff
@@ -0,0 +1,23 @@
...
+#define DATAOBJECTS_GEANTEVENTENTRY_H
+
+//Include files
+#include "GaudiKernel/DataObject.h"
+#include "G4Event.hh"
+
+class GAUDI_API GeantEventEntry : public DataObject {
+ public:
+  GeantEventEntry() : m_event(0){};
+   virtual ~GeantEventEntry() {if (m_event!=0) delete m_event;};
```
sbinet: this should read:

```c++
virtual ~GeantEventEntry() {
delete m_event;
m_event = nullptr;
};
```
otherwise, double-deletes will happen.
also, probably one should use a smart pointer of some kind very early on in FCC-SW.

hegner: Very true!

Thanks for catching,
  Benedikt

On 17 Mar 2015, at 14:45, Sebastien Binet <notifications@github.com<mailto:notifications@github.com>> wrote:


In DataObjects/DataObjects/GeantEventEntry.h<https://github.com/HEP-FCC/FCCSW/pull/25#discussion_r26572861>:

> @@ -0,0 +1,23 @@
> +#ifndef DATAOBJECTS_GEANTEVENTENTRY_H
> +#define DATAOBJECTS_GEANTEVENTENTRY_H
> +
> +//Include files
> +#include "GaudiKernel/DataObject.h"
> +#include "G4Event.hh"
> +
> +class GAUDI_API GeantEventEntry : public DataObject {
> + public:
> +  GeantEventEntry() : m_event(0){};
> +   virtual ~GeantEventEntry() {if (m_event!=0) delete m_event;};


this should read:

virtual ~GeantEventEntry() {
delete m_event;
m_event = nullptr;
};

otherwise, double-deletes will happen.
also, probably one should use a smart pointer of some kind very early on in FCC-SW.

Simulation/src/Geant4Simulation.h
```diff
@@ -0,0 +1,36 @@
...
+  virtual StatusCode finalize();
+private:
+  /* /// Pointer to Geant4 geometry */
+  /*G4VUser.... * m_g4detector; */
+  /// Handle for the G4Event to be read
+  DataObjectHandle<GeantEventEntry> m_eventhandle;
+  /// Handle for the "reconstructed" to be written
+  DataObjectHandle<ParticleCollection> m_recphandle;
+
+  G4RunManager* m_runManager;
```
sbinet: same comment than above: we should use smart pointers, not bare pointers.

DataObjects/DataObjects/GeantEventEntry.h
sbinet: also, this should probably be moved to a `.cxx` file otherwise there won't be any anchor for the vtable (and this will bloat the executable)

zaborowska:
I've updated the pull request.
After the talk with Benedikt it turned out the separate hepmc->g4 conversion is useless.

hegner:
Thanks. Against which Geant4 version did you develop here?

zaborowska:
At this level I checked both G4 9.6 and G4 10.1, but the general idea is to have it working with 10.1.

hegner:
I merged it, but had to fix for changes in the latest Geant4 preview. CLHEP units now need to be explicitly imported.

PR #28, opened 2015-03-26
modify pythia config

clementhelsens:
change the lhe file in the pythia config file


clementhelsens:
self approve this pull request, small bug fix

PR #29, opened 2015-03-30
Geant simulation

zaborowska:
Geant simulation:
1. Detector TEMPORARY read from gdml file (FCCDetectorConstruction, FCCFullDetector.gdml to be changed/deleted) - to be replaced with the DD4hep detector
2. Physics list includes Fast SImulation Models
3. Event read in Gaudi from HepMC event-by-event and passed to Geant Run Manager
4. Output TEMPORARY saved with Geant tools (FCCOutput, FCCPrimaryParticleInformation, to be changed/deleted) - to be replaced with the EDM

Current stage does not include Atlfast: just simple parametrisation, pT-dependent.


GeantFast/src/FCCSmearer.cc
```diff
@@ -0,0 +1,75 @@
...
+#include "G4UnitsTable.hh"
+#include "G4SystemOfUnits.hh"
+#include "G4TransportationManager.hh"
+#include "G4FieldManager.hh"
+#include "G4UniformMagField.hh"
+#include <ctime>
+
+
+//....oooOO0OOooo........oooOO0OOooo........oooOO0OOooo........oooOO0OOooo......
+FCCSmearer* FCCSmearer::fFCCSmearer = 0;
```
sbinet: this should read nullptr_t

GeantFast/src/FCCSmearer.cc
```diff
@@ -0,0 +1,75 @@
...
+   fRandomGauss = new CLHEP::RandGauss(fRandomEngine);
+}
+//....oooOO0OOooo........oooOO0OOooo........oooOO0OOooo........oooOO0OOooo......
+FCCSmearer::~FCCSmearer()
+{}
+
+//....oooOO0OOooo........oooOO0OOooo........oooOO0OOooo........oooOO0OOooo......
+FCCSmearer* FCCSmearer::Instance()
+{
+   if(!fFCCSmearer)
```
sbinet: that's racy and a one-time leak which pollutes valgrind reports. prefer this (still non-thread-safe):

```c++
static FCCSmearer smearer;
fFCCSmearer = &smearer;
return fFCCSmearer;
```

GeantFast/GeantFast/FCCSmearer.hh
```diff
@@ -0,0 +1,72 @@
...
+   A default constructor.
+*/
+   FCCSmearer();
+   ~FCCSmearer();
+
+private:
+/**
+   A pointer to FCCSmearer object.
+*/
+   static FCCSmearer* fFCCSmearer;
```
sbinet: this doesn't look like a good idea, in a multithreaded world a la GaudiHive: a singleton with a PRNG as data member with no mutex lock won't fly in MT.

GeantFast/src/FCCSmearer.cc
hegner: actually nullptr ;-)

GeantFast/GeantFast/FCCSmearer.hh
hegner: correct. The entire G4 interface ATM is rather shaky. It's work in progress though.

sbinet: of course, I was just being "cap'tain obvious" here (and not dismissive of anything in that PR), playing by the PR rules.

PR #32, opened 2015-04-10
Pyroot

clementhelsens:
Not albe to load ROOT 5.38 in python 2.6, need python 2.7


cbernet:
Hi Clement, 

What do you mean? I can perfectly import ROOT in python with the current settings. 
Please contact me so that we can sort that our. In the meanwhile I will close. 

C

PR #34, opened 2015-05-15
Pythiainterface

clementhelsens:
Adding counter in pythiainterface to abort run after N flawed events
N being controled in the main.cmnd by
Main:timesAllowErrors = N

This is needed because sometimes Pythia failed to hadronize an event out of many and we don't want to stop the event loop for that.


cbernet:
Hi Clement, 

I'm not completely sure about the logic. 
Don't you want to allow a given number of aborts in a raw? For instance, if the user sets it to 1, pythia would be allowed to abort once in a raw. If there are two aborts in a raw, then the processing would stop. 

The problem with the existing logic is that the user needs to estimate the expected number of aborts depending on the number of events to be processed, wich is far from being practical. 

To solve this, you should set iabort to 0 as soon as an event succeeds. 

Two other comments: 

- can't we test the status of pythia after an abort to skip only the events which fail the fragmentation(while other kinds of abort would lead to a termination of the process) 

- instead of the cout at L58, can't you issue a warning using the logging system? I don't know how to do that though (@hegner?)

Cheers, 

Colin

clementhelsens:
Hi Colin, 

thanks for the suggestion, I haven't considered doing it in a raw, which would be a more robust test of a potential misconfiguration. But from my experience, Pythia is not supposed to failed very often, so having an overall counter of how many events have failed might be enough. But I don't have a strong preference, so I will change my pull request to count failed events in a raw

For your first comment: 
the Pythia->next() command returns a boolean which could be false for various reasons (failed hadronization, energy not conserved, reached the end of les houche event file etc...) So unless we change pythia, we do not have handles on what the problem is. As a first step, and given the manpower, I would recommand to modify Pyhtia.

For your second command: yes, if Benedikt could either tell me how to do it or modify my request that woudl be great!

hegner:
Hi Clement, Colin,

that's fairly easy. The algorithm has a few methods allowing that. You can replace the 
  std::cout << ..
with 
  warning() << ...

Below all the methods of that kind at your disposal.

Cheers,
  Benedikt

MsgStream & 	always () const
 	shortcut for the method msgStream ( MSG::ALWAYS )
 
MsgStream & 	fatal () const
 	shortcut for the method msgStream ( MSG::FATAL )
 
MsgStream & 	err () const
 	shortcut for the method msgStream ( MSG::ERROR )
 
MsgStream & 	error () const
 	shortcut for the method msgStream ( MSG::ERROR )
 
MsgStream & 	warning () const
 	shortcut for the method msgStream ( MSG::WARNING )
 
MsgStream & 	info () const
 	shortcut for the method msgStream ( MSG::INFO )
 
MsgStream & 	debug () const
 	shortcut for the method msgStream ( MSG::DEBUG )
 
MsgStream & 	verbose () const
 	shortcut for the method msgStream ( MSG::VERBOSE )
 
MsgStream & 	msg () const
 	shortcut for the method msgStream ( MSG::INFO )

clementhelsens:
The iAbort counter is now set to 0 once Pythia->next() succeed.
I also used the method Benedikt proposed and fix some bugs in the includes (Generation->Generators)

PR #35, opened 2015-05-19
Delphessim 190515

degrutto:
fix event handle in delphes


cbernet:
Hi Michele, 
I'll let Benedikt review this PR when he's back as I've not followed that. 
Just one thing for the future: you should use 

git commit --amend 

if you want to add to an existing commit instead of duplicating the commit message. If, on the contrary, these commits correspond to specific developments, you should use different, explicit commit messages. 

Cheers! 

Colin

hegner:
Hi Michele,

giving it a quick look it seems much closer, yes. The basic structure is correct. However, can you confirm you did test it beyond the 1-event case and compared the results? 

For quality assurance - could you replace the cout statements by proper Gaudi logging? And remove the progress bar, which I believe shouldn't do a nice job anyhow with the interleaving framework messages.

Thanks,
  Benedikt

degrutto:
Hi Benedikt,

Yes I did a test on 10000 events and I get the very same results for delphes standalone and within fccsw (on the common dataformat... still the dataformat of delphes is more advanced, but this is another issue)

ok I can remove the progress bar and replace the cout,

please wait for a new commit on this same branch
cheers,
     Michele

degrutto:
done, at least I believe it is what you requested,
please let me know if that is still not the case,

cheers and thanks

  Michele

PR #38, opened 2015-10-07
Cleaning

clementhelsens:
Hi there, 

I removed some pythia configuration files and cleaned a bit the naming for the pythia_standard and pythia_lhe examples.
Also added 1000 lhe events to the config directory.

Clement


clementhelsens:
I self approve this pull request :)

PR #41, opened 2015-11-04
Update FCChh_DelphesCard_WithDipole_v00.tcl

fmoortga:
Bugfixes for photons, muons and electrons.
Update of the muon pt cut and resolution.

### DIFF
--- a/config/FCChh_DelphesCard_WithDipole_v00.tcl
+++ b/config/FCChh_DelphesCard_WithDipole_v00.tcl
@@ -37,7 +37,7 @@ TowerMerger
 EFlowMerger
 
 PhotonEfficiency
-PhotonEnergySmearing
+#PhotonEnergySmearing
 
 MuonIsolation
 ElectronIsolation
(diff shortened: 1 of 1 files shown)


clementhelsens:
Hi Filip, 

is it supposed to be the final version 0 of the card we should use?

Cheers,
Clement

fmoortga:
Hi Clement,

yes, this is the final one.
Before everyone starts using it, we still need to double check that the physics output looks reasonable.
So it would be good if someone (Zbynek or yourself) could run with this card over a background sample (e.g. Z+jets) to make sure that the distributions look fine.

Cheers, Filip

config/FCChh_DelphesCard_WithDipole_v00.tcl
```diff
@@ -37,7 +37,7 @@ TowerMerger
 EFlowMerger
 
 PhotonEfficiency
-PhotonEnergySmearing
+#PhotonEnergySmearing
```
clementhelsens: Delphes does not like comments in the ExecutionPath.
to make it usable one needs to remove '#PhotonEnergySmearing'

fmoortga: Ah, thanks. Can you remove it (or uncomment it, doesn't really matter) or should I make a new pull request?

cbernet: Hi Filip, It's probably easier if you do it. In fact, you can just append commits to your branch patch-1 and they will appear here automatically. So that's the same PR, you're just modifying it on the fly. Cheers & thanks!

fmoortga: Thanks Colin, got it!
Cheers, Filip

clementhelsens:
I ran over some events and look at some basics plots that look good to me, but more detailed cross-checks are needed.
I will merge this pull request and continue the validation with a demonstrator to be build together with  @drasal

PR #44, opened 2015-11-30
Update externals + code refactoring

hegner:
This (work-in-progress) pull request updates to new versions of ROOT 6, Geant4, DD4hep, and alike. It as well simplifies the environment setting.
Please be aware that this pull-request already includes the PR #39 of Anna Zaborowska.


clementhelsens:
Thanks Benedkit, 
I went quickly through the code and I see that in some directories both the .src and .hh are present. Is this intended?

hegner:
Where we have plugins/components it is intended. The headers for these we do not want to and shouldn't expose for further inclusion.

FWCore/FWCore/DataWrapper.h
```diff
@@ -23,8 +25,7 @@ class GAUDI_API DataWrapper : public DataWrapperBase {
   const T* getData() {return m_data;} 
   void setData(T* data) {m_data = data;}
 
-  virtual albers::CollectionBase* collectionBase(){return dynamic_cast<albers::CollectionBase*>(m_data); }
-
+  virtual albers::CollectionBase* collectionBase(){if (std::is_base_of<albers::CollectionBase,T>::value) return reinterpret_cast<albers::CollectionBase*>(m_data); return nullptr;}
```
jlingema: This line got a bit hard to read with the change. Is it possible to add a few newlines? Regarding the inlining: Not sure if this still counts as a trivial implementation :-)

jlingema:
Regarding Clement's comment: In the long run we would like to split packages in a way where they only contain either components / plugins or entities that need to be exposed to other packages. The `src` directory should also have sub-directories `components` and `lib`. To keep track of this (which is not part of this PR), we created a ticket here:
https://sft.its.cern.ch/jira/browse/FCC-44

FWCore/FWCore/DataWrapper.h
jlingema: As a second thought: There should maybe be some documentation that this may return a nullptr.

hegner:
Joschka, I did the cleanup you were suggesting.

clementhelsens:
In the README file, we still request the user to 
export FCCSW=$FCCBASE/FCCSW

But $FCCBASE is not defined, and I don't think $FCCSW Environment varialbe is used. Shall we remove this instruction as well?

cbernet:
Hi Clement, 

You can use github to give comments on specific lines.
Could you please do that? Now I don’t really see what you’re talking about :-) 

Colin

clementhelsens:
sure, will do!

Le 5 janv. 2016 à 12:42, Colin Bernet <notifications@github.com<mailto:notifications@github.com>> a écrit :

Hi Clement,

You can use github to give comments on specific lines.
Could you please do that? Now I don’t really see what you’re talking about :-)

Colin

README.md
```diff
@@ -36,21 +36,6 @@ Installation
 
 Log in to lxplus SLC6. Please note that these instructions might not work on another SLC6 machine. Instructions to compile on a mac are in preparation. 
 
-Prepare a global software zone for FCC:
-
-    mkdir FCC
-    cd FCC
-    export FCCBASE=$PWD
```
clementhelsens: Here we remove export "FCCBASE=$PWD"

README.md
```diff
@@ -58,15 +43,13 @@ Clone the FCCSW repository: **replace \<username\> by your github username**
```
clementhelsens: here we keep "export FCCSW=$FCCBASE/FCCSW" while "export FCCBASE=$PWD" has been previously removed
I think the environment variable FCCSW is not needed anymore

cbernet: nice catch! 

The FCCSW env var is used in the tutorial. 
https://twiki.cern.ch/twiki/bin/view/FCC/FccSoftwareFramework

I have no strong opinion on what to do with this variable, but it's important to make sure that our tutorials work.

clementhelsens: Then I guess the easiest solution is to redefine 
"export FCCSW=$FCCBASE/FCCSW" 
to
"export FCCSW=$PWD/FCCSW"

PR #48, opened 2016-01-05
Fixing examples and updating README

jlingema:
Some example configs were missing the vertex output for the `HepMCConverter` which was re-added. Also updating the README according to Clement's comments in PR #44.

@clementhelsens can you let me know if this fixes the issues you were seeing? Thanks!

PS: Sorry for the white-space strips.

### DIFF
--- a/Examples/config/delphes_simulation.py
+++ b/Examples/config/delphes_simulation.py
@@ -10,15 +10,15 @@ delphessimulation = DelphesSimulation("Simulation", filename="/afs/cern.ch/user/
 #delphessimulation = DelphesSimulation("Simulation", filename="example_MyPythia.dat" , detectorcard = "Simulation/examples/delphes_card_FCC_basic.tcl" , outputcollections =  [ "ElectronEnergySmearing/electrons", "MuonMomentumSmearing/muons",  "JetEnergyScale/jets",  "MissingET/momentum", "ScalarHT/energy" ] )
 
 # we shouls add b-jets and track inpact parameters, fixme
-delphessimulation.Outputs.genparticles.Path = "genparticles"
-delphessimulation.Outputs.particles.Path = "particles"
-delphessimulation.Outputs.partons.Path = "partons"
-delphessimulation.Outputs.muons.Path = "muons"
-delphessimulation.Outputs.electrons.Path = "electrons"
-delphessimulation.Outputs.photons.Path = "photons"
-delphessimulation.Outputs.jets.Path = "jets"
-delphessimulation.Outputs.mets.Path = "met"
-delphessimulation.Outputs.hts.Path = "ht"
+delphessimulation.DataOutputs.genparticles.Path = "genparticles"
+delphessimulation.DataOutputs.particles.Path = "particles"
+delphessimulation.DataOutputs.partons.Path = "partons"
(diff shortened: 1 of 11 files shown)


jlingema:
@clementhelsens, I just saw that some configs were not migrated to `DataOutputs`, I'll fix that.

clementhelsens:
Hi, 

the example "config/simple_workflow.py" seems to work, but there is still Error messages from TCling::
----------------------
Error in <TCling::RegisterModule>: cannot find dictionary module AlbersDict_rdict.pcm
Error in <TCling::RegisterModule>: cannot find dictionary module ExampleDict_rdict.pcm
----------------------
@drasal I tried the config/PythiaDelphes_config.py example and it failed, could you please look into it?

drasal:
Hi Clement and Joschka,

I have cross-checked the Pythia + Delphes chain and found the following:

* Pythia without Delphes module runs correctly
* Once the Delphes module is processed several errors regarding TCling and missing *_dict.pcm appears. 
* I have commented out the whole code withing the Delphes module to find out if a specific part is concerned or something more serious causes the troubles. The latter seems to be the problem.

If there is a way how to compile the whole code against ROOT version<6.0, I would like to try it out. Do  you have some hints how to proceed on lxplus machine? I think that some incompatibility between Delphes module and ROOT >=6.0 might cause these troubles. I'm not an expert, but as far as I understand it, the Delphes code uses ROOT streamer to stream out/in TClass objects. All that is based on TCling & dictionaries generated by it. The missing dictionaries  *_dict.pcm seems to be pointing to this problem.

jlingema:
Hi Zbynek, 

Thanks for having a look at this. Indeed it is a problem of incompatibility. However, it is not the missing PCM dictionaries, but rather that the existing ones for Delphes don't match what the module is trying to stream. I suspect that the dicts were compiled with a different ROOT version(?).

The problem can be solved by regenerating the dictionaries. Could you please check whether it works for you with the ones I just generated by doing:
```
export LD_LIBRARY_PATH=/afs/cern.ch/user/j/jlingema/public/forZbynek:$LD_LIBRARY_PATH
./run gaudirun.py config/PythiaDelphes_config.py
```

jlingema:
Hi,

I added a workaround for the shadow warnings and recompiled Delphes in the FCC release area, there is a small complication to avoid breaking FCCSW for anybody who cloned before this PR [0]

From my point of view this PR could be merged as soon as somebody checked the Delphes example works for them.

Cheers,
Joschka

[0] Since I recompiled Delphes using CMake, the directory structure of Delphes changed. The needed changes are in the PR, but I left the old installation to ensure that FCCSW still compiles for anybody who uses the version prior to this PR.

zaborowska:
Hi all,
I checked delphes example and it seems to work now, no Cling errors any more nor any other.

Can you, Joschka, just add output for "genvertices" to config/simple_pythia.py ? (as in 
https://github.com/jlingema/FCCSW/commit/9018baa36ad687ed04b705fe97fbf5ce827858b8)

jlingema:
Thanks for testing. I added the gen-vertices to the config in the last commit.

clementhelsens:
merge this PR!

PR #54, opened 2016-02-05
DevDelphes - updated output to FCC-EDM (based on new PODIO library)

drasal:
The Delphes simulation module provides now a correct conversion of Delphes internal objects to FCC-EDM objects (based on PODIO) and relates them simultaneously to MC truth information. The definition, which Delphes arrays are automatically translated to FCC-EDM, is given through a GAUDI Pythia+Delphes configuration file. Currently, muons, electrons, charged & neutral hadrons, photons, jets and METS objects are saved together with a full MC particle tree. More details can be found either in a Sim/SimDelphesInterface/src/DelphesSimulation.h header file or config/PythiaDelphes_config.py file.


hegner:
test this please

hegner:
Could you add a test/test scenario for this adaptor to the pull request? Thanks!

drasal:
Dear Benedikt, it has been intensively tested with the real HH data sample.

hegner:
@drasal : we would need some test we can run automatically on this comparing delphes and what your translation did. don't want to validate that manually next time we do changes. 

In any case thanks for the pull request. :-) I would have a few general remarks/questions. One is the usage of bare pointers. Please reduce that as much as possible. Why were they needed? Then there are a few strange parts of code where you statically cast iterators. And please instead of uncommenting old code parts, please remove them. I will add a few more concrete remarks later today.

config/PythiaDelphes_config.py
```diff
@@ -36,17 +50,25 @@
...
+## provide the same type of particle with different level of processing ...
+## Needed due to required output to FCC-EDM!
+delphesMuonsOutArray    ="MuonMomentumSmearing/muons"
+delphesElectronsOutArray="ElectronEnergySmearing/electrons"
+delphesChargedOutArray  ="ChargedHadronMomentumSmearing/chargedHadrons"
+delphesNeutralOutArray  ="Hcal/eflowNeutralHadrons"
+delphesPhotonsOutArray  ="PhotonEfficiency/photons"
+delphesJetsOutArray     ="JetEnergyScale/jets"
+delphesMETsOutArray     ="MissingET/momentum"
+delphesSHTsOutArray     ="ScalarHT/energy"
```
hegner: Why do you define these variables if you use their value only once each? Just simplify.

Sim/SimDelphesInterface/src/ParticleStatus.h
```diff
@@ -0,0 +1,23 @@
...
+ *
+ *  Lists possible statuses of simulated or reconstructed particles
+ *
+ *  @author: Z. Drasal (CERN)
+ *
+ */
+#ifndef INCLUDE_PARTICLE_STATUS_H_
+#define INCLUDE_PARTICLE_STATUS_H_
+
+namespace ParticleStatus {
```
hegner: I'd use a typed enum here.

Sim/SimDelphesInterface/src/ParticleStatus.h
```diff
@@ -0,0 +1,23 @@
+/** @class ParticleStatus
+ *
+ *  Lists possible statuses of simulated or reconstructed particles
+ *
+ *  @author: Z. Drasal (CERN)
+ *
+ */
+#ifndef INCLUDE_PARTICLE_STATUS_H_
```
hegner: see
  http://fccsw.web.cern.ch/fccsw/FCCSW/md_doc__cpp_coding_style_guidelines.html#include-guard

Sim/SimDelphesInterface/src/DelphesSimulation.h
```diff
@@ -70,45 +77,66 @@ class DelphesSimulation: public GaudiAlgorithm {
...
+
+  DataHandle<fcc::ParticleMCParticleAssociationCollection>  m_handleRecMuonsToMC;     // Relation between muons & MC particle
+  DataHandle<fcc::ParticleMCParticleAssociationCollection>  m_handleRecElectronsToMC; // Relation between electrons & MC particle
+  DataHandle<fcc::ParticleMCParticleAssociationCollection>  m_handleRecChargedToMC;   // Relation between charged hadrons & MC particle
+  DataHandle<fcc::ParticleMCParticleAssociationCollection>  m_handleRecNeutralToMC;   // Relation between neutral hadrons & MC particle
+  DataHandle<fcc::ParticleMCParticleAssociationCollection>  m_handleRecPhotonsToMC;   // Relation between photons & MC particle
+  DataHandle<fcc::GenJetParticleAssociationCollection>      m_handleRecJetsToMC;      // Relation between jets & MC particle
+
+  // Convert internal Delphes objects to FCC EDM objects - UPDATE
+  void ConvertMCParticles(const TObjArray* Input,    fcc::MCParticleCollection* colMCParticles,
```
hegner: passing references would be better. Is there a reason why this wouldn't do?

drasal:
Sure, I understand that there needs to be automatic testing. But this time it was unavoidable, due to tests I managed to understand how to make it at all. I will think of automatic testing once it's done, I agree that manual testing is not optimal at all.

You're pointing at:
auto bareJet      = new fcc::BareJet();
well, referencing provided zero values. There might be a trick how to do it without, but I didn't find it. Please, send me a correct example, how to do it, I will try it and correct the code accordingly. 

There is no old code inside. These are the testing outputs.

Sim/SimDelphesInterface/src/DelphesSimulation.h
```diff
@@ -70,45 +77,66 @@ class DelphesSimulation: public GaudiAlgorithm {
...
-  std::string              m_DelphesCard;
-
-  // Output collections to be write out
-  std::vector<std::string> m_outCollections;
+  std::string            m_DelphesCard;
 
   // Delphes
   Delphes               *m_Delphes ;
   DelphesFactory        *m_DelphesFactory ;
   DelphesExtHepMCReader *m_HepMCReader ;
```
hegner: will the Delphes HepMCReader eventually disappear? The idea was to pass HepMC into Delphes from our side... e.g. required if we want to add pile-up ourselves.

config/PythiaDelphes_config.py
drasal: Because, the user might want different output from Delphes:
For photons one might get a photon before isolation procedure or after:
PhotonEfficiency/photons or PhotonIsolation/photons 

The same for muons, electrons, etc.

Sim/SimDelphesInterface/src/ParticleStatus.h
drasal: This is for testing purposes, one should use bit encoder/decoder instead. But yes, I agree enum is probably better.

config/PythiaDelphes_config.py
hegner: What I meant is that you can as well do this around line 70ff.

Sim/SimDelphesInterface/src/ParticleStatus.h
drasal: My mistake, I'm sorry.

Sim/SimDelphesInterface/src/DelphesSimulation.cpp
```diff
@@ -343,78 +369,452 @@ StatusCode DelphesSimulation::finalize() {
...
-  for(int j = 0; j < Input->GetEntries(); ++j) {
+void DelphesSimulation::ConvertTracks(const TObjArray*  Input,
+                                      fcc::ParticleCollection*   colParticles,
+                                      fcc::MCParticleCollection* colMCParticles,
+                                      fcc::ParticleMCParticleAssociationCollection* ascColParticlesToMC)
+{
+  Candidate* cand = nullptr;
+  for(int j=0; j<Input->GetEntries(); j++) {
 
     cand = static_cast<Candidate *>(Input->At(j));
```
hegner: I don't understand why you need these static_casts here and at other places. What's the return-type before casting?

Sim/SimDelphesInterface/src/DelphesSimulation.cpp
```diff
@@ -343,78 +369,452 @@ StatusCode DelphesSimulation::finalize() {
...
+      }
+    }
+    else {
+      barePart->Bits = ParticleStatus::Unmatched;
+      particle.Core(*barePart);
+      warning() << "Can't build relation from Electron/Muon/ChHadron to MC particle!" << std::endl;
+    }
+    if (barePart!=nullptr) delete barePart;
+
+    // Test print
```
hegner: Please remove these lines if they don't serve any purpose any more. Or make it a debug output.

Sim/SimDelphesInterface/src/DelphesSimulation.cpp
```diff
@@ -343,78 +369,452 @@ StatusCode DelphesSimulation::finalize() {
...
+
+    // Reference to MC - Delphes holds references to all objects related to the Photon object, several relations might exist for gammas
+    std::set<int> idRefMCPart; // Avoid double counting when referencingh MC particles
+
+    Candidate* refCand  = nullptr;
+    Candidate* ref2Cand = nullptr;
+
+    TIter it1(cand->GetCandidates());
+    it1.Reset();
+    while((refCand = static_cast<Candidate*>(it1.Next()))) {
```
hegner: here you could just try using iterators; isn't that supported?

Sim/SimDelphesInterface/src/DelphesSimulation.cpp
```diff
@@ -343,78 +369,452 @@ StatusCode DelphesSimulation::finalize() {
...
+    // Reference to MC - Delphes holds references to all objects related to the Jet object,
+    // several relations might exist -> find "recursively" in a tree history a MC particle
+    std::set<int> idRefMCPart; // Avoid double counting when referencingh MC particles
+
+    Candidate* refCand  = nullptr;
+    Candidate* ref2Cand = nullptr;
+    Candidate* ref3Cand = nullptr;
+    Candidate* ref4Cand = nullptr;
+
+    TIter it1(cand->GetCandidates());
```
hegner: Could you explain why four levels of loops are necessary? The code is not intuitive.

Sim/SimDelphesInterface/src/DelphesSimulation.cpp
```diff
@@ -343,78 +369,452 @@ StatusCode DelphesSimulation::finalize() {
...
+    // Test print
+    /*double totSimE = 0;*/
+
+    for (auto id : idRefMCPart) {
+
+      auto relation = ascColJetsToMC->create();
+      relation.Jet(jet);
+      relation.Particle(colMCParticles->at(id));
+
+      // Test print
```
hegner: remove deprecated code. see above.

Sim/SimDelphesInterface/src/DelphesSimulation.cpp
```diff
@@ -343,78 +369,452 @@ StatusCode DelphesSimulation::finalize() {
...
+                           relation.Particle().Core().P4.Pz*relation.Particle().Core().P4.Pz +
+                           relation.Particle().Core().P4.Mass*relation.Particle().Core().P4.Mass);
+      totSimE += simE;
+      std::cout << " RefId: "    << setw(3)  << id+1
+                << " Rel E: "    << setw(10) << totSimE << " " << simE << " <-> " << recE << std::endl;*/
+    }
+    // Test print
+    /*std::cout << std::endl;*/
+
+    if (bareJet!=nullptr) delete bareJet;
```
hegner: that makes me believe that you either want to use a unique_ptr or just something on the stack.

Sim/SimDelphesInterface/src/DelphesSimulation.cpp
```diff
@@ -343,78 +369,452 @@ StatusCode DelphesSimulation::finalize() {
...
+
+    auto bareJet      = new fcc::BareJet();
+    bareJet->Area     = -1;
+    bareJet->P4.Px    = cand->Momentum.Px();
+    bareJet->P4.Py    = cand->Momentum.Py();
+    bareJet->P4.Pz    = cand->Momentum.Pz();
+    bareJet->P4.Mass  = cand->Mass;
+    jet.Core(*bareJet);
+
+    // Test print
```
hegner: see above; remove deprecated code

Sim/SimDelphesInterface/src/DelphesSimulation.cpp
```diff
@@ -244,84 +265,89 @@ StatusCode DelphesSimulation::execute() {
...
-  m_genStablesHandle.put(genStableParticles);
-  for ( unsigned int l=0; l<m_outCollections.size() ; l++){
-    if ( m_outCollections[l].find("muons")    != std::string::npos ) m_recMuonHandle.put(recMuons);
-    if ( m_outCollections[l].find("electrons")!= std::string::npos ) m_recElecHandle.put(recElectrons);
-    if ( m_outCollections[l].find("photons")  != std::string::npos ) m_recPhotHandle.put(recPhotons);
-    if ( m_outCollections[l].find("jets")     != std::string::npos ) m_recJetsHandle.put(recJets);
-    if ( m_outCollections[l].find("mets")     != std::string::npos ) m_recMETHandle.put(recMETs);
-    if ( m_outCollections[l].find("hts")      != std::string::npos ) m_recHTSHandle.put(recHTs);
-  }*/
+  auto genParticles     = new fcc::MCParticleCollection();
```
jlingema: Not sure if this actually leaks memory, but I think it would be better to use `fcc::MCParticleCollection *genParticles(nullptr);` (same below)

Sim/SimDelphesInterface/src/DelphesSimulation.cpp
```diff
@@ -343,78 +369,452 @@ StatusCode DelphesSimulation::finalize() {
...
+    // Test print
+    /*double totSimE = 0;*/
+
+    for (auto id : idRefMCPart) {
+
+      auto relation = ascColParticlesToMC->create();
+      relation.Rec(particle);
+      relation.Sim(colMCParticles->at(id));
+
+      // Test print
```
hegner: see above; remove deprecated code

Sim/SimDelphesInterface/src/DelphesSimulation.cpp
```diff
@@ -343,78 +369,452 @@ StatusCode DelphesSimulation::finalize() {
...
+    barePart->P4.Py    = cand->Momentum.Py();
+    barePart->P4.Pz    = cand->Momentum.Pz();
+    barePart->P4.Mass  = cand->Momentum.M();
+    barePart->Charge   = cand->Charge;
+    barePart->Vertex.X = cand->Position.X();
+    barePart->Vertex.Y = cand->Position.Y();
+    barePart->Vertex.Z = cand->Position.Z();
+    particle.Core(*barePart);
+
+    // Test print
```
hegner: see above; remove deprecated code

Sim/SimDelphesInterface/src/DelphesSimulation.cpp
jlingema: Alternatively you could also try to just get the collections from the handles (`auto genParticles = m_handleGenParticles.get();`)

Sim/SimDelphesInterface/src/DelphesSimulation.cpp
```diff
@@ -244,84 +265,89 @@ StatusCode DelphesSimulation::execute() {
...
+  m_chargedOutArray  = m_Delphes->ImportArray(m_DelphesChargedArrayName.c_str());   // "ChargedHadronMomentumSmearing/chargedHadrons"
+  m_neutralOutArray  = m_Delphes->ImportArray(m_DelphesNeutralArrayName.c_str());   // "Hcal/eflowNeutralHadrons"
+  m_photonOutArray   = m_Delphes->ImportArray(m_DelphesPhotonsArrayName.c_str());   // "PhotonEfficiency/photons" / "PhotonIsolation/photons"
+  m_jetOutArray      = m_Delphes->ImportArray(m_DelphesJetsArrayName.c_str());      // "JetEnergyScale/jets"
+  m_metOutArray      = m_Delphes->ImportArray(m_DelphesMETsArrayName.c_str());      // "MissingET/momentum"
+  m_shtOutArray      = m_Delphes->ImportArray(m_DelphesSHTsArrayName.c_str());      // "ScalarHT/energy"
+
+  if (m_muonOutArray    ==nullptr) warning () << "Can't save Delphes muon array: "
+                                              << m_DelphesMuonsArrayName
+                                              << " to FCCEDM. Doesn't exist!!!";
```
jlingema: maybe add a `endmsg()` here?

Sim/SimDelphesInterface/src/DelphesSimulation.cpp
```diff
@@ -343,78 +369,452 @@ StatusCode DelphesSimulation::finalize() {
...
+    while((refCand = static_cast<Candidate*>(it1.Next()))) {
+
+      TIter it2(refCand->GetCandidates());
+      it2.Reset();
+      while((ref2Cand = static_cast<Candidate*>(it2.Next()))) {
+
+        int id = ref2Cand->GetUniqueID()-1;
+        if (id<colMCParticles->size()) idRefMCPart.insert(id);
+        else {
+          warning() << "Can't build one of the relations from Photon/NHadron to MC particle!" << std::endl;
```
hegner: Why a warning? Is that something one can ignore? Having that fail - is that a feature of the input data or the algorithm?

Sim/SimDelphesInterface/src/DelphesSimulation.cpp
drasal: TObject

jlingema:
Regarding the testing output: If this might be needed in the future, why don't you use the debug stream instead of commenting it out?

Sim/SimDelphesInterface/src/DelphesSimulation.cpp
drasal: Ok. I will put them as debug output. Good hint.

Sim/SimDelphesInterface/src/DelphesSimulation.cpp
drasal: This is the official Delphes code, so I followed it.

Sim/SimDelphesInterface/src/DelphesSimulation.cpp
hegner: I see. If you don't need cand outside the loop, please put the definition inside that scope.

Sim/SimDelphesInterface/src/DelphesSimulation.cpp
drasal: Well, that's very difficult to answer as some particles in Delphes point directly to the MC particle, some through 1 particle, some through 2 particles and some through 3 particles. That's why, the deep mechanism is hidden within Delphes and I didn't have time to go in detail through the Delphes code. But I understand that this is not a clear solution. I may put there a recursive calling ...

Sim/SimDelphesInterface/src/DelphesSimulation.cpp
hegner: Better follow the standard protocol. Supported here: 
  https://root.cern.ch/doc/master/classTIter.html

you may try whether it already works with
  http://www.cprogramming.com/c++11/c++11-ranged-for-loop.html

Sim/SimDelphesInterface/src/DelphesSimulation.cpp
drasal: Can you be more specific here. Sorry, my experience is not that large.

Sim/SimDelphesInterface/src/DelphesSimulation.h
drasal: I do delete the pointer in the finalize() method...

Sim/SimDelphesInterface/src/DelphesSimulation.h
drasal: Delphes returns a pointer to TObjArray, so I've been working with that. As for the collection, I do create collection dynamically, so then naturally I'm passing them by a pointer. I don't see any possible leak here. If the preferred way is to pass them by reference, it can surely be done.

config/PythiaDelphes_config.py
drasal: Ok.

Sim/SimDelphesInterface/src/DelphesSimulation.h
hegner: Nah. The point was to have the delphes interface not reading the HepMC file directly, but getting the HepMC event from the FCC framework; in that case the Delphes internal HepMCReader is not needed, right?

Sim/SimDelphesInterface/src/DelphesSimulation.cpp
drasal: Alright.

Sim/SimDelphesInterface/src/DelphesSimulation.cpp
drasal: Ok. Good point.

Sim/SimDelphesInterface/src/DelphesSimulation.cpp
hegner: You mean you "experimentally" found out that 3 levels is enough? You are going through a tree here until you reach the MC particles. Maybe the work of Alice can help.

Sim/SimDelphesInterface/src/DelphesSimulation.cpp
drasal: Ok.

Sim/SimDelphesInterface/src/DelphesSimulation.cpp
drasal: I will put debug() in there. Is there several debug levels defined or just one debug level?

Sim/SimDelphesInterface/src/DelphesSimulation.cpp
drasal: I'm sorry, I didn't get the point. What is wrong about this part of the code. I'm using TIterators... Can you give me a short example.

Sim/SimDelphesInterface/src/DelphesSimulation.cpp
drasal: This is warning that the algorithm is not working correctly. According to the Delphes code, it should be alright. But if they change it, we have no chance to spot it without a warning line.

hegner: If it fails, it fails. So it should stop and not warn.

Sim/SimDelphesInterface/src/DelphesSimulation.cpp
drasal: No, the Delphes code works until this level of deepness, but one can do it recursively (so something similar what Alice did). In some cases there is direct link, in some cases in goes through related particle, etc. I will think of improving the code.

Sim/SimDelphesInterface/src/DelphesSimulation.h
drasal: No, it's not needed then.

Sim/SimDelphesInterface/src/DelphesSimulation.cpp
hegner: see 
http://www.cprogramming.com/tutorial/stl/iterators.html
("The STL approach")
plus the second link above.

Sim/SimDelphesInterface/src/DelphesSimulation.cpp
jlingema: Only one debug level, see [gaudi-doxygen](http://proj-gaudi.web.cern.ch/proj-gaudi/releases/v26r3/doxygen/d6/d6b/class_gaudi_common.html#ab3f3a23fca6a3375ac5f99c61ae8590d)

Sim/SimDelphesInterface/src/DelphesSimulation.cpp
drasal: Ok.

Sim/SimDelphesInterface/src/DelphesSimulation.h
```diff
@@ -3,10 +3,14 @@
...
+//#include "datamodel/GenJetCollection.h"
 #include "datamodel/ParticleCollection.h"
+#include "datamodel/ParticleMCParticleAssociationCollection.h"
 #include "datamodel/GenJetCollection.h"
+#include "datamodel/GenJetParticleAssociationCollection.h"
 #include "datamodel/METCollection.h"
-#include "datamodel/MCParticleCollection.h"
 #include "TFile.h"
 #include "FWCore/DataHandle.h"
 #include "HepMC/GenEvent.h"
```
jlingema: I think several of these could just be forward declarations and then only included in the implementation file.

Sim/SimDelphesInterface/src/DelphesSimulation.cpp
drasal: So, you want to use STL iterators instead of ROOT iterators?

hegner: That's just how code could look like, e.g. (untested for syntax errors):
`
for(auto cand = it1.begin(); cand != it1.end(); ++it1 ) {... }
`

Sim/SimDelphesInterface/src/DelphesSimulation.cpp
jlingema: see http://de.cppreference.com/w/cpp/memory/unique_ptr and https://github.com/HEP-FCC/FCCSW/blob/master/doc/GeneralCppGuidelines.md#poiners-and-references

Sim/SimDelphesInterface/src/DelphesSimulation.cpp
drasal: Ok. I know the STL iterators, I just clearly didn't understand what is wrong. So, preferred are STL iterators  ... I will follow them.

Sim/SimDelphesInterface/src/DelphesSimulation.cpp
drasal: Thanks for the comment. I still don't get what is wrong here. Usage of raw pointers is wrong or something more deep?

Sim/SimDelphesInterface/src/DelphesSimulation.h
drasal: Sure, I will test it.

Sim/SimDelphesInterface/src/DelphesSimulation.h
drasal: So, should I read the HepMC through FCC framework?

hegner: eventually yes.

Sim/SimDelphesInterface/src/DelphesSimulation.cpp
jlingema: I think the comment was not to use STL iterators (as you're using a ROOT array here), but rather to use STL-like syntax, i.e. have a look at `Begin` and `End` of ROOT container. The reasoning is that this is likely easier to understand for somebody who is not familiar with ROOT, as it looks more like what you would do with STL containers.

Sim/SimDelphesInterface/src/DelphesSimulation.cpp
jlingema: The point is that this is more error prone than using either stack allocation or "smart pointers" (e.g. `unique_ptr`). Since the lifetime of the objects you are creating here is defined by the scope, it makes your life easier not to use pointers. While you now remembered to delete them, somebody modifying the code in the future (copy pasting something) might not.

Sim/SimDelphesInterface/src/DelphesSimulation.h
```diff
@@ -70,45 +77,66 @@ class DelphesSimulation: public GaudiAlgorithm {
...
-  void  ConvertHT(      TObjArray *  Input, fcc::ParticleCollection * coll );
-
 private:
 
   // Delphes detector card to be read in
-  std::string              m_DelphesCard;
-
-  // Output collections to be write out
-  std::vector<std::string> m_outCollections;
+  std::string            m_DelphesCard;
```
jlingema: This could also be done in another iteration, but since I see that you did some doxygen documentation of the class :-). Could you add some comments for the data and function members? Some names are not completely clear (e.g. the *OutArrays at first glance might look like your output not the Delphes output).

Generation/src/PythiaInterface.cpp
```diff
@@ -14,38 +17,48 @@ PythiaInterface::PythiaInterface(const std::string& name, ISvcLocator* svcLoc):
   GaudiAlgorithm(name, svcLoc), m_pythia( nullptr ), m_parfile()
 {
   declareProperty("Filename", m_parfile="", "Name of the Pythia parameter file to read");
-  declareOutput("hepmc", m_hepmchandle);
+  declareOutput(  "hepmc"   , m_hepmchandle);
+
+  m_nAbort = 0;
```
jlingema: Please initialise all members in the initialiser list.

Generation/src/PythiaInterface.cpp
```diff
@@ -53,55 +66,89 @@ StatusCode PythiaInterface::execute() {
...
-                //<< " " << m_pythia->event[i].mother1()
-                //<< " " << m_pythia->event[i].mother2()
-                //<< " " << m_pythia->event[i].daughter1()
-                //<< " " << m_pythia->event[i].daughter2()
-                << " " << m_pythia->event[i].px()
-                << " " << m_pythia->event[i].py()
-                << " " << m_pythia->event[i].pz() 
-                << " " << m_pythia->event[i].e()
-                << " " << m_pythia->event[i].m()
+      std::cout << "PythiaInterface : "
```
jlingema: I guess this should rather go to debug than cout?

Generation/src/PythiaInterface.cpp
```diff
@@ -53,55 +66,89 @@ StatusCode PythiaInterface::execute() {
...
+
+      int daughterID      = -1;
+      int daughterIDRange = 0;
+      if ((*ipart)->end_vertex()!=nullptr) {
+
+        daughterID      = (*((*ipart)->end_vertex()->particles_out_const_begin()))->barcode();
+        daughterIDRange = (*ipart)->end_vertex()->particles_out_size() -1;
+      }
+
+      std::cout << "Pythia HepMC: "
```
jlingema: Please change this to debug stream..

Generation/src/PythiaInterface.cpp
```diff
@@ -53,55 +66,89 @@ StatusCode PythiaInterface::execute() {
...
-              << " " << (*ipart)->momentum().pz() 
-              << " " << (*ipart)->momentum().e()
-              << " " << (*ipart)->momentum().m()
-              << std::endl;
-  */
 
+  // Print debug: HepMC event info
+  if (msgLevel() <= MSG::DEBUG) {
+
+    for (HepMC::GenEvent::particle_iterator ipart = theEvent->particles_begin() ;
```
jlingema: not very important as it is in the debugging clause, but this would be nicer:
```c++
for (auto ipart = theEvent->particles_begin(), end = theEvent->particles_end();
      ipart != end; ++ipart) {
```

Generation/src/PythiaInterface.h
```diff
@@ -22,14 +27,19 @@ class PythiaInterface: public GaudiAlgorithm {
   virtual StatusCode finalize();
 
 private:
+
   // Pythia8 engine
   Pythia8::Pythia * m_pythia;
-  /// the name of the Pythia parameter input file
+
+  // Name of Pythia configuration input file
```
jlingema: Why did you change the doxygen comment to a normal comment? Please re-add the third "/".

Sim/SimDelphesInterface/src/DelphesSimulation.cpp
```diff
@@ -1,5 +1,33 @@
...
+#include "datamodel/GenJetCollection.h"
+#include "datamodel/GenJetParticleAssociationCollection.h"
+#include "datamodel/METCollection.h"
+
+// ROOT
+#include "TFile.h"
+#include "TObjArray.h"
+#include "TStopwatch.h"
+
 using namespace std;
```
jlingema: please re-frain from using `using namespace std`. Instead add explicit statements for the constructs you want to use from `std`. E.g. `using std::cout`.

Sim/SimDelphesInterface/src/DelphesSimulation.cpp
```diff
@@ -21,32 +48,49 @@ GaudiAlgorithm(name, svcLoc) ,
...
-   declareOutput("genPartons"        , m_genPartonsHandle);
-   declareOutput("genStableParticles", m_genStablesHandle);
-   declareOutput("recMuons"          , m_recMuonHandle);
-   declareOutput("recElectrons"      , m_recElecHandle);
-   declareOutput("recPhotons"        , m_recPhotHandle);
-   declareOutput("recJets"           , m_recJetsHandle);
-   declareOutput("recMETs"           , m_recMETHandle);
-   declareOutput("recHTs"            , m_recHTSHandle);
 
    m_stablePartOutArray = nullptr;
```
jlingema: Please initialise all members in the member initialiser list.

Sim/SimDelphesInterface/src/DelphesSimulation.cpp
```diff
@@ -88,28 +132,28 @@ StatusCode DelphesSimulation::initialize() {
     }
   }
 
-  // Read Delphes configuration card
+  // Read Delphes configuration card (deleted by finalize())
   m_confReader = new ExRootConfReader;
```
jlingema: Why don't you use smart-pointers here? Same question applies to the other raw pointer members.

Sim/SimDelphesInterface/src/DelphesSimulation.cpp
```diff
@@ -165,67 +209,75 @@ StatusCode DelphesSimulation::execute() {
...
-      int daughterID              = -1;
-      int daughterIDRange         = 0;
-      if ((*ipart)->production_vertex()!=nullptr) {
-
-        motherID      = (*((*ipart)->production_vertex()->particles_in_const_begin()))->barcode();//(*((*ipart)->production_vertex()->particles_begin()))->barcode();
-        motherIDRange = (*ipart)->production_vertex()->particles_in_size() -1;
+    // Print debug: HepMC event info
+    if (msgLevel() <= MSG::DEBUG) {
+
+      for(auto ipart=hepMCEvent->particles_begin(); ipart!=hepMCEvent->particles_end(); ++ipart) {
```
jlingema: Same comment as for the for-loop in PythiaInterface.

Sim/SimDelphesInterface/src/DelphesSimulation.cpp
```diff
@@ -165,67 +209,75 @@ StatusCode DelphesSimulation::execute() {
...
+          motherID      = (*((*ipart)->production_vertex()->particles_in_const_begin()))->barcode();
+          motherIDRange = (*ipart)->production_vertex()->particles_in_size() -1;
+        }
+        if ((*ipart)->end_vertex()!=nullptr) {
+
+          daughterID      = (*((*ipart)->end_vertex()->particles_out_const_begin()))->barcode();
+          daughterIDRange = (*ipart)->end_vertex()->particles_out_size() -1;
+        }
+
+        std::cout << "Delphes HepMC: "
```
jlingema: Probably should also go to debug stream?

Sim/SimDelphesInterface/src/DelphesSimulation.cpp
```diff
@@ -165,67 +209,75 @@ StatusCode DelphesSimulation::execute() {
...
+                  << " Mothers: "  << setw(3)  << motherID   << " -> " << setw(3) << motherID  +motherIDRange
+                  << " Daughters: "<< setw(3)  << daughterID << " -> " << setw(3) << daughterID+daughterIDRange
+                  << " Stat: "     << setw(2)  << (*ipart)->status()
+                  << " Px: "       << setprecision(2) << setw(9) << (*ipart)->momentum().px()
+                  << " Py: "       << setprecision(2) << setw(9) << (*ipart)->momentum().py()
+                  << " Pz: "       << setprecision(2) << setw(9) << (*ipart)->momentum().pz()
+                  << " E: "        << setprecision(2) << setw(9) << (*ipart)->momentum().e()
+                  << " M: "        << setprecision(2) << setw(9) << (*ipart)->momentum().m();
+        if ((*ipart)->production_vertex()!=nullptr) {
+        std::cout << " Vx: "       << setprecision(2) << setw(9) << (*ipart)->production_vertex()->position().x()
```
jlingema: Increase indentation level.

Sim/SimDelphesInterface/src/DelphesSimulation.cpp
jlingema: Now you can actually use the createAndPut function here :-)

Sim/SimDelphesInterface/src/DelphesSimulation.cpp
```diff
@@ -343,78 +400,514 @@ StatusCode DelphesSimulation::finalize() {
     m_outRootFile->Close();
     if (m_outRootFile){delete m_outRootFile; m_outRootFile = nullptr;}
   }
-
+  
   info() << "Exiting Delphes..." << endmsg;
-
+  
   // Clear memory
   if (m_HepMCReader) {delete m_HepMCReader; m_HepMCReader = nullptr; } // Releases also the memory allocated by inHepMCFile
```
jlingema: Add a comparison to nullptr or use smart-pointers.

Sim/SimDelphesInterface/src/DelphesSimulation.cpp
```diff
@@ -343,78 +400,514 @@ StatusCode DelphesSimulation::finalize() {
...
+
+  m_vecPartProdVtxIDDecVtxID.resize(Input->GetEntries());
+  for(int j=0; j<Input->GetEntries(); j++) {
+    m_vecPartProdVtxIDDecVtxID[j].first  = -1;
+    m_vecPartProdVtxIDDecVtxID[j].second = -1;
+  }
 
-  Candidate * cand;
-  for(int j = 0; j < Input->GetEntries(); ++j) {
+  Candidate* cand = nullptr;
```
jlingema: Looks to me that the scope of cand could be decreased to the for-loop below.

drasal:
I have updated the code following Joschka's recommendations:

* PythiaInterface: 
 - no raw pointer 
 - added doxygen comments 
 - moved std::cout print to GAUDI standard print outputs (debug() mostly)

* DelphesSimulation 
 - removed some of raw pointers (still some remaining, need to study in more detail Delphes memory management, as the pointers are passed as input parameters to Delphes functions, before I transfer them to smart pointer quantities)
 - added doxygen comments
 - moved std::cout print to GAUDI standard print outputs (debug() mostly)

Remaining issues:
 - Delphes related pointers need to changed to smart pointers (the next step)
 - MC particle tree for some particles broken if LHE file from Madgraph used (under progress)
 - Tag info missing in FCC-EDM & Delphes output (the next step)

hegner:
@drasal - the tests fail because of

>  PYTHIA Abort from Pythia::init: Les Houches Event File not found 

can you address that?

drasal:
It seems that the automatic test failed because it couldn't find the options/Pythia_LHEinput.cmd, which is in the options directory. Is the directory, where the test is run set to FCCSW directory?

drasal:
@hegner When I run it locally at LXPLUS the same procedure is without any problem.

vvolkl:
Hi Zbynek,
I had a similar problem restructuring the Generation package. The tests are run from the build directory, but you may try to set `WORKING_DIRECTORY ${PROJECT_SOURCE_DIR}` in `gaudi_add_tests` in `CMakeLists.txt`

drasal:
Hi, thanks a lot for the hint. But I checked that in the main CMakeLists.txt it's set like that ...

jlingema:
I believe it's due to this line:
https://github.com/drasal/FCCSW/blob/devDelphes/options/Pythia_LHEinput.cmd#L42
The file is not in the repository. Also remember that for the tests you'll need to specify filenames relative to the FCCSW top directory.

drasal:
Thank you Joschka! I will fix it.

Sim/SimDelphesInterface/src/DelphesSimulation.cpp
```diff
@@ -341,80 +409,516 @@ StatusCode DelphesSimulation::finalize() {
...
+  //static int depth = 0;
+  //depth++;
+
+  // Warning - no MC relation found
+  if (jetPart->GetCandidates()->GetEntries()==0) {
+
+    warning() << "Can't build one of the relations from Jet to MC particle!" << std::endl;
+  }
+  // Relation can be found
+  else for (auto itCand=jetPart->GetCandidates()->begin(); itCand!=jetPart->GetCandidates()->end(); ++itCand) {
```
jlingema: Please add curly braces here for the `else` here.

jlingema: Also please make sure that the for-loops use this slightly nicer syntax that avoids multiple function calls:
```c++
for (auto itCand=jetPart->GetCandidates()->begin(), end = jetPart->GetCandidates()->end(); itCand != end; ++itCand)
```

Sim/SimDelphesInterface/src/DelphesSimulation.cpp
```diff
@@ -18,37 +43,54 @@ GaudiAlgorithm(name, svcLoc) ,
...
+
+  declareOutput("recMuonsToMC"      , m_handleRecMuonsToMC);
+  declareOutput("recElectronsToMC"  , m_handleRecElectronsToMC);
+  declareOutput("recChargedToMC"    , m_handleRecChargedToMC);
+  declareOutput("recNeutralToMC"    , m_handleRecNeutralToMC);
+  declareOutput("recPhotonsToMC"    , m_handleRecPhotonsToMC);
+  declareOutput("recJetsToMC"       , m_handleRecJetsToMC);
+
+  // All variables truly initialized in the initialize() method (for pedantic C++ reasons also in a constructor)
+  m_stablePartOutArray = nullptr;
```
jlingema: Please initialise all members in the member initialiser list.

Sim/SimDelphesInterface/src/DelphesSimulation.cpp
jlingema: Comment is still valid.

Sim/SimDelphesInterface/src/DelphesSimulation.cpp
drasal: They are all initialized in the member initialiser list. Look below.

Sim/SimDelphesInterface/src/DelphesSimulation.cpp
drasal: GetCandidates returns a pointer, so there is no CPU overhead for direct call. Curl braces will be added.

Sim/SimDelphesInterface/src/DelphesSimulation.cpp
jlingema: The member initialiser list is above this point in the code :-).. I.e. where you initialise a lot of the other members.

drasal: Ah Ok.

jlingema:
Thanks a lot for implementing the changes! Merging this now.

PR #58, opened 2016-02-09
Adding create and put to data handle

jlingema:
Adding the possibility to directly retrieve the pointer to a new collection from the DataHandle. This makes for a slightly nicer interface, compare:
```c++
auto coll = new CollectionType();
// do something with coll
myHandle.put(coll);
```
and 
```c++
auto coll = myHandle.createAndPut();
// do something with coll
```


hegner:
very good! could you add this to one of the tests?

jlingema:
I changed the FastSim example, which is run in the tests, to use the new interface.

PR #59, opened 2016-02-09
Summary documentation of examples

zaborowska:
There are two main changes:
- moving `PodioOutput` and `KeepDropSwitch` from Examples/ (where they didn't really belong) to FWCore/
- renaming `PodioWrite` to `CreateSampleJet` as it is just a simple example of an algorithm creating the EDM collection

Additionally:
- obsolete options files were removed
- many comments inside our main examples were added
- list of examples was created

To consult:
1. Lists come in 2 styles:
[standard list](https://github.com/zaborowska/FCCSW/blob/examples/doc/Examples.md)
[table](https://github.com/zaborowska/FCCSW/blob/examples/doc/ExamplesTable.md)
with ~ the same content, but I couldn't decide which I like better.
The standard list of examples looks pretty long and boring and I prefer in general tables (nicer comparison), but it doesn't look that nice on Github... but with doxygen it is OK.
2. Since we have directory Examples/ maybe we should consider moving the main dir options/ there?

Please let me know what you think


jlingema:
Hi Anna, 

This is very nice! I personally agree that `Examples/options` would be a better place. Could you maybe move the summaries into `Examples/doc`? I also would prefer the table.

Cheers,
Joschka

vvolkl:
Nice!  -- I also agree with you on the table.

hegner:
Great. Table is nicer indeed! Agree with all of Joschka's comments.

vvolkl:
I just noticed that doxygen creates pages for the options-scripts, e.g. http://fccsw.web.cern.ch/fccsw/FCCSW/geant__fastsim_8py.html.  Maybe we can include some doxygen style comments, so they show up there?

drasal:
I've been discussing this with Anna ...

I've noticed that the config directory has been renamed to options directory and Anna asked me if all its content can come to Examples/options. Personally, I would copy only the config files which are connected with the examples. The rest I'd keep in a general directory options or config or make the options / config directory in Sim directory / Generation directory, dependent on what they relate to. Let me stress that some files are official FCC-hh configuration files (FCC-hh Delphes card etc.), so they should be kept in some official directory. Any opinion on that?

jlingema:
Hi Zbynek,

I agree to some extend. I also think that some of the configurations currently in the top level would belong more into package directories. However, having a central folder that has examples for newcomers sounds appealing to me and `Examples/options` sounds like a good choice for that.

To your point about the Delphes cards: I think they actually don't belong into the options (or config) directory at all. They maybe could be put into the Delphes package in a folder like "data"? Alternatively they could also naturally belong into the FCC-hh repo (if that is still planned to be used) and could additionally be put into the afs-sw area?

zaborowska:
Maybe we can create `data/` directory in every relevant package?
* pythia*.cmd, *lhe files could go to Generation/data
* Delphes cards to `Sim/SimDelphes/data` ?

And the rest, so all *py job options files can go to `Examples/options` ?

hegner:
sounds OK to me. Even though these data are somehow examples as well. But Examples should just contain *complete* ones that can be run.

jlingema:
If they are only examples wouldn't `Examples/data` make sense?

hegner:
Can you if you know how to do it add an extra test_examples target? So that one can selectively run all those?

hegner:
@jlingema - I would anticipate that sooner or later these examples will use "real" MC cards. And then you don't want to have a dummy one and a real one and people being confused about double maintenance.

drasal:
I agree that these configurations are indeed examples, but in an example directory I'd be looking for an example module and its config file. So, if the configuration files are related to official modules, I'd keep them in an options directory, which will be directly under Generation, Simulation etc. As for the data, I'm not sure here. I agree that they should be kept in an independent FCC-hh directory, which doesn't exist now. So, it's on your decision.

zaborowska:
Evolution of directory `options/`:
* Delphes cards were moved to `Sim/SimDelphesInterface/data`
* Pythia (and LHE) moved to `Generation/data` (also ParticleTable.txt is now there)
* py script to display geometry moved to `DetectorDescription/tools`
* the rest are Gaudi options files and they were moved to `Examples/options`, except for delphes example that is now in `Sim/SimDelphesInterface/options/`

The [table with examples](https://github.com/zaborowska/FCCSW/blob/examples/Examples/doc/ExamplesTable.md) stayed, but now in `Examples/doc`.

Also, every example is also put in an appropriate CMakeLists.txt as a test.

drasal:
Thanks Anna, I will update the Delphes config files and Pythia file soon with the update of Delphes to the newly created directories.

zaborowska:
Oh, @drasal it is not a problem, leave them as they are now and once my PR will be merged (after your), those files will be moved. I will resolve any potential conflicts.

drasal:
My commit will come after yours, so it's better if I follow the new structure. It will be less work for you.

hegner:
@drasal - we won't merge Anna's work before yours. So just leave that part of yours as is.

hegner:
@zaborowska - can you fix the conflicts after the last merge? Thanks.

zaborowska:
@hegner - done and checks passed

PR #60, opened 2016-02-09
Fix compile time warnings

jlingema:
I went through the warnings and fixed most of them (those that are within FCCSW and not in the Delphes package (waiting for @drasal's changes)). Also, I did a couple of changes to make the code compile with clang (needed for static checks with scan-build). 

Here are some things that I'm not very happy with and would like some input for (hence the WIP tag):
- Some functions are currently complete studs (they do nothing) should I just remove those?

Cheers,
Joschka

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -13,15 +13,10 @@ find_package(Delphes)
 gaudi_project(FCCSW HEAD
               USE Gaudi v26r4 )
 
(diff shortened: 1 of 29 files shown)


jlingema:
OK, after speaking with @jhrdinka, the functions that are empty can be ignored and will be deleted by her in another PR.

hegner:
retest this please

PR #61, opened 2016-02-10
Fix PodioOutput to demangle namespace correctly

jlingema:
Thanks a lot to @vvolkl for spotting this. Due to the addition of namespaces, the output collections were not saved correctly because the names were not correctly demangled. 

I added a quick fix, but as the comment at the beginning of the function states, we should improve how we get the class names.


jlingema:
I did a bit of googling and found a solution that seems to be rather robust (also supporting multiple namespaces). It uses the ABI directly, I tested with clang and gcc and both work. The interface is very C-ish, unfortunately. I leave it up to you what you prefer. The alternative implementation can be found here (tried my best to document why it's done the way it is done):
https://github.com/jlingema/FCCSW/blob/demangling/Examples/src/PodioOutput.cpp#L75

PR #62, opened 2016-02-11
clean up repository from old Trackingsoftware

jhrdinka:
Cleaned up repository from old tracking software (including reconstruction geometry) and its tests to be replaced by the ATS tracking software in near future.


jlingema:
Hi Julia,

I think you may have forgotten to push some commits?

Cheers,
Joschka

PR #67, opened 2016-02-15
Updating coding style and adding script for format checking

jlingema:
Hi,

I updated the coding style document so the folder structure now includes tests and its sub-directories as well as data. I also noted that the indentation of member initialiser lists should be 4 not 2 spaces.

There is also a new script that does some basic format checking based on a format file (for now in my [SAS fork](https://github.com/jlingema/SAS/blob/master/config/FCCSW-clang-format.yaml) and in the afs FCCSW area). If we want to add the format file to the FCCSW repo, let me know.

Cheers,
Joschka


hegner:
Thanks. Could you add another build executing this once a day?

PR #68, opened 2016-02-15
Set length unit to mm

jlingema:
Some length units were set to cm, now it should consistently be mm.


hegner:
@jlingema - I don't like that approach. What about a single units header with typedefs?

jlingema:
@hegner OK, sure. I'll open a new PR tomorrow.

PR #71, opened 2016-02-18
Fix podio output for collections with relations

jlingema:
Hi all,

This fixes a segv we had when reading collections that have relations. Additionally, I cleaned up the code a bit and made variable names a bit more explicit. 

Note, that currently this only works when the collections are "put" before creating the relations (a corresponding fix for PODIO is already implemented but needs to be pushed to the HEP-FCC fork).

Cheers,
Joschka

### DIFF
--- a/FWCore/components/PodioOutput.h
+++ b/FWCore/components/PodioOutput.h
@@ -42,9 +42,9 @@ private:
   /// The actual ROOT file
   std::unique_ptr<TFile> m_file;
   /// The tree to be filled with collections
-  std::unique_ptr<TTree> m_datatree;
+  TTree* m_datatree;
   /// The tree to be filled with meta data
-  std::unique_ptr<TTree> m_metadatatree;
+  TTree* m_metadatatree;
   /// The stored collections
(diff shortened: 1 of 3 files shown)


FWCore/components/PodioOutput.h
```diff
@@ -19,22 +19,33 @@ class PodioOutput : public GaudiAlgorithm {
...
   KeepDropSwitch m_switch;
+  /// Needed for collection ID table
   PodioDataSvc* m_podioDataSvc;
-  TFile* m_file;
-  TTree* m_datatree;
-  TTree* m_metadatatree;
+  /// The actual ROOT file
+  std::unique_ptr<TFile> m_file;
+  /// The tree to be filled with collections
+  std::unique_ptr<TTree> m_datatree;
```
hegner: The TTrees should be owned by ROOT after adding them to the file. Thus you cannot do an automatic cleanup w/ smart pointers.

jlingema:
I'll also update the init.sh once we have tagged and installed new versions of FCC-EDM and Podio. I'll keep the [WIP] prefix until that's done.

PR #74, opened 2016-02-26
Adding tools and infrastructure for magnetic field in G4

Ada15764:
Adding the possibility of setting a constant magnetic field in G4SimSvc. The implementation in the Svc is the same as for the detector geometry tool, which means that a G4Magneticfield tool has to be explicitly instantiated in the job option for the thing to work. if the tool is not explicitly instantiated, the Svc will return FAILURE on initialization. it would be good to have a procedure that allows:
a) to check if the ToolHandle was initialized to avoid retrieving the tool if it was not (e.g. no field)
b) to set a default tool to be invoked in case of no explicit instantiation (e.g. field from a field map as default property)
I would appreciate any suggestion about points a) and b) above


jlingema:
Hi Andrea,

You can set a default when declaring the tool, so that would take care of b):
```
    // arguments are: ToolHandle, DefaultToolName, createIfNotInstantiated
    declarePrivateTool(m_magneticFieldTool, "ConstantMagneticFieldTool", true);
```
If you add this, it also takes care of the failing tests.

I guess if we want a), we'd need another tool that represents no field. 

Cheers,
Joschka

Ada15764:
Thank you Joschka, I'm going to try that out

andrea

jlingema:
Hi again,

OK, that fixed all tests but the one running fast simulation. Anna is currently investigating where it's failing when the magnetic field is introduced. We'd like to wait with the merge until that is understood and fixed.

In the meantime I had a look at the code. Could you prefix the new class names with `SimG4` (except for the interface) and split `Sim/SimG4Components/src/ConstantMagneticFieldTool.h` into two files each containing one class definition?

Thanks,
Joschka

Sim/SimG4Components/src/ConstantMagneticFieldTool.cpp
```diff
@@ -0,0 +1,128 @@
...
+}
+
+
+//=============================================================================
+// Finalize 
+//=============================================================================
+StatusCode ConstantMagneticFieldTool::finalize( )
+{
+	info()<<" ConstantMagneticFieldTool::finalize( ) "<<endmsg;
+	StatusCode sc = GaudiTool::initialize( ) ;
```
jlingema: Can you change this to `finalize()`?

jlingema:
retest this please

PR #76, opened 2016-03-02
DevDelphes - flavour tags, isolation tags saved on output, added smart pointers

drasal:
New features in DelphesSimulation:

* Newly generated jets saved on output with relation to MC particles.
* Reconstructed jets use Jet class now (instead of GenJet class), on output the relation from jet to jet constituents saved (instead of to MC particles).
* The flavour tag and respective associations for generated jets and reconstructed jets are saved in FCC-EDM format.
* Similarly, the isolation tags for electrons, muons and photons and their respective associations are saved in FCC-EDM format.
* Added smart pointers where possible
* Adapted the configuration file PythiaDelphes_config.py

New features in PythiaInterface:

* Added automatic random seed settings to Pythia_standard.cmd


jlingema:
test this please

jlingema:
retest this please

jlingema:
retest this please

jlingema:
Hi Zbynek,

Thanks for the update. Have you tried using the FCCSW HepMC reader, are there problems with it for your use case?

Looking at the config: Could make sense to set the defaults for the outputs where you declare them in your algorithm. 

Cheers,
Joschka

drasal:
Joschka, 

thanks for the comment. The main purpose of the DelphesExtHepMCReader is not to read the HepMC event from file, but to read all the data from HepMC container to the internal structures of Delphes, see:

isEventReady = m_HepMCReader->ReadEventFromStore(hepMCEvent, m_Delphes->GetFactory(), m_allPartOutArray, m_stablePartOutArray, m_partonOutArray);

 In principal, this class can also read the HepMC event from file and save its content to internal structures of Delphes, but that's currently an option, not the main stream. So, one inevitably needs such a class to fill in all Delphes containers & cross-check that everything is set-up and available for each event. Finally, it provides the output to Delphes natural structures, which is again not a main stream. On the other hand, it's a standard Delphes output and may be needed for testing purposes.

As for your second comment, it seems that it might make the config file slightly more readable, so good point!

jlingema:
Hi Zbynek,

I had a look at the classes. If I understand things correctly, the reader fills the objects it gets from the generated event into TObjArrays via its class members. Since we want to be able to run algorithms in parallel, this is a bit error prone. Also, in my opinion, this is an unnecessary step (GenEvent -> TObjArray -> EDM instead of just doing GenEvent -> EDM, the checks / relations could also just be done there). 

If you're busy with other stuff, I can have a look and change this once you're finished from your point of view. That reminds me: What's the status of tests we can run to make sure everything works as expected?

Cheers,
Joschka

drasal:
Hi Joschka,

yes, you understand this correctly. Nevertheless, Delphes internally works with the TObjArray, so you might need to somehowe fill the internal Delphes objects. Anyway, feel free to change it, if you find a better solution. But please, keep in mind, that it has to remain consistent with the Delphes internal structures. 

As for the tests, I will come back to them after the Rome Week, I need to focus on tracker results now. Unfortunately, for the Delphes module. But I will keep this in my head.

 Cheers and thanks,
 Zbynek

drasal:
To be compatible with the changes in Delphes card by Michele, I have adapted the b-tagging output. In addition, I have added the c-tagging output. That wasn't possible before the changes in the Delphes card. Currently the B Tagging module uses bits: 0 (for b-tags), 1 (for c-tags) -> the same procedure is used when extracting this info to FCC-EDM.

jlingema:
retest this please

PR #78, opened 2016-03-03
Rename readout so it is recognized by SaveHits-Tool.

vvolkl:
The current implementation of the tracker endcaps does not result in saved hits when running the simulation examples. It should conform with the conventions documented [here](https://github.com/vvolkl/FCCSW/blob/tracker_readout_fix/Sim/doc/Geant4fullsim.md#note).  As a fix, I just rename the readout.


jlingema:
test this please

PR #81, opened 2016-03-04
update delphes cards to comply with Delphes-3.3.2

selvaggi:


drasal:
Dear Michele,

can you, please, comment on the following? Why have you chosen as an input array for the ElectronFilter module the following:

InputArray HCal/eflowTracks

I'd somehow naturally expect the ECal/eflowTracks input when filtering electrons, but I might be wrong...

 Thanks a lot!

selvaggi:
Hi Zbynek,

you are right that electrons are not affected by HCal so in practice you can put either as input to the ElectronFilter.
However, I think it is clearer this way, because the final e-flow collection is produced by the HCal.
This said, feel free to change it if you think it is clearer the other way.

Cheers,
Michele


On 04 Mar 2016, at 16:15, Zbynek Drasal <notifications@github.com<mailto:notifications@github.com>>
 wrote:


Dear Michele,

can you, please, comment on the following? Why have you chosen as an input array for the ElectronFilter module the following:

InputArray HCal/eflowTracks

I'd somehow naturally expect the ECal/eflowTracks input when filtering electrons, but I might be wrong...

Thanks a lot!

drasal:
Thank you, Michele!

hegner:
Why not replacing the existing files instead of adding new ones? That may create confusion on which configs are supposed to work and which are outdated.

selvaggi:
Ok, I have replaced the original cards.

clementhelsens:
revert the merging. card needs to be validated first by @fmoortga and Heather Gray

PR #84, opened 2016-03-08
Add easy way to compile against dbg libraries.

jlingema:
- Add switch to compile against dbg libraries:
```
export BUILDTYPE=Debug
source init.sh
make
```
- Switch automatically enables debug build of FCCSW itself

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -1,7 +1,7 @@
 ################################################################################
 #
 # Generic Makefile to simplify the use of CMake projects
-# ------------------------------------------------------ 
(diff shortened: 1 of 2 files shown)


jlingema:
Could this be due to merged #81? Looks like the Delphes module is not very happy with the input:
```
DelphesSimulation   ERROR can't access input list 'ElectronEnergySmearing/electrons' in module 'Delphes'
```

jlingema:
retest this please

PR #86, opened 2016-03-10
Cleanup of FCChh calorimeter xml files, revision of dimensions

faltovaj:


hegner:
test this please

clementhelsens:
we don't have other tests than compiling the code.
not sure what would be the best tests to be done there when we modify/add new detectors...
as we discussed already, we might want to compare with a benchmark.

hegner:
@clementhelsens - that was the command to the buildbot ;-)

jlingema:
@clementhelsens we also run the full sim with HCal, so the tests would've spotted if the xmls were renamed but the config not updated accordingly :).

clementhelsens:
ok I see, but as soon as we add a new detector, we have to build a full sim test accordingly, no?

jlingema:
Yes, of course...

zaborowska:
All tests regarding full sim should go to Sim/SimG4Components/tests/,
following what Joschka described here: [FCCSW/doc/AddingTestsToFCCSW.md](https://github.com/HEP-FCC/FCCSW/blob/master/doc/AddingTestsToFCCSW.md).

So, in principle, for any new detector a test may be created (based on [existing ones](https://github.com/HEP-FCC/FCCSW/blob/master/Sim/SimG4Components/tests/geant_fullsim_hcal.py)).

clementhelsens:
thanks! everything is clear now!

jlingema:
retest this please

DetectorDescription/Detectors/compact/FCChh_ECalDefinition.xml
```diff
@@ -9,3 +9,10 @@
     <passive_layers thickness="5*mm" material="Lead" />
   </calorimeter>
 </detector>
+
+  <readouts>
+    <readout name="BarECal_Readout">
+      <segmentation type="CartesianGridXY" grid_size_x="0.05*mm" grid_size_y="0.05*mm"/>
+      <id>system:5,layer:50,x:32:-16,y:-16</id>
```
jlingema: Why do you redefine the readout here? Also: The ID is a 64 bit number of which you define 50 to be taken for layer identification and 32 for segmentation, which doesn't work out. I suspect that this could have to do with the time out.

DetectorDescription/Detectors/compact/FCChh_ECalBarrel_Mockup.xml
```diff
@@ -0,0 +1,53 @@
...
+    <constant name="world_z" value="world_size"/>
+
+    <!---- ECAL dimensions ----->    
+    <constant name="BarECal_rmin" value="2600*mm"/>
+    <constant name="BarECal_rmax" value="3500*mm"/>
+    <constant name="BarECal_dz" value="4000*mm"/>
+    
+    <constant name="EMBarrel_rmin" value="2700*mm"/>
+    <constant name="EMBarrel_rmax" value="3400*mm"/>
+    <constant name="EMBarrel_dz" value="3900*mm"/>
```
jlingema: Could you add a comment why the two sets of dimensions are needed?

faltovaj:
I removed the readout from FCChh_ECalDefinition.xml, that was a part of "forgotten" code. Sorry about that.
The first set of dimensions is for cryostat, the second one for the actual calorimeter. I guess you would like to have better names, right?

jlingema:
Hi,

Yes.. Either better names or a comment in the code which variables refer to what...

Cheers,
Joschka

faltovaj:
I added a comment. Hope it is OK now.

jlingema:
retest this please

DetectorDescription/Detectors/compact/FCChh_ECalBarrel_Mockup.xml
```diff
@@ -0,0 +1,55 @@
...
+    <comment>ECal very conceptual design</comment>
+  </info>
+
+  <define>
+    <!-- World volume --->
+    <constant name="world_size" value="30*m"/>
+    <constant name="world_x" value="world_size"/>
+    <constant name="world_y" value="world_size"/>
+    <constant name="world_z" value="world_size"/>
+
```
clementhelsens: Why is the world volume defined in the detector xml?
I think it is better to define it once in a "master" xml as already done (FCChh_DectMaster.xml)

faltovaj:
More readout fixes, hope the tests will succeed finally!
To Clement: I can remove the world definition, you just have to decide. It was commented out first, Joschka told me to add it.

jlingema:
What I meant was to rather remove the code than to comment it out :-)

faltovaj:
OK, I see :) I will do it in a minute!

jlingema:
retest this please

jlingema:
OK, the only test failing is Delphes as mentioned in PR #89. So we can merge this one.

jlingema:
retest this please

PR #87, opened 2016-03-10
Python config classes for different pileup scenarios.

vvolkl:
Add an easier way to set pileup parameters globally without recompiling.


jlingema:
test this please

Generation/options/pileup_hepmcreader.py
```diff
@@ -6,18 +6,12 @@
...
 
-# assuming mm
-smeartool = FlatSmearVertex(
-    xVertexMin = -0.5,
-    xVertexMax = 0.5,
-    yVertexMin = -0.5,
-    yVertexMax = 0.5,
-    zVertexMin = -35,
-    zVertexMax = 35 )
+from FCCPileupScenarios import FCCPhase1Pileup
```
hegner: no imports in the middle of code.

Generation/python/FCCPileupScenarios.py
```diff
@@ -0,0 +1,28 @@
...
+    """Beam Parameters Common to all FCC scenarios"""
+    def __init__(self, name):
+        self._name = name
+        self.VertexSmearingConfig = {
+          "xVertexMin": -0.5,
+          "xVertexMax": 0.5,
+          "yVertexMin": -0.5,
+          "yVertexMax": 0.5,
+          "zVertexMin": -35,
+          "zVertexMax": 35
```
hegner: Could you add a reference to the paper where you took the parameters from?

Generation/options/pileup_hepmcreader.py
```diff
@@ -6,18 +6,12 @@
...
+pileupscenario = FCCPhase1Pileup('PileupScenario')
+print pileupscenario.PileupToolConfig
```
hegner: I'd not have printouts per default.

vvolkl: Oh. That was for debugging and slipped through the cracks.

vvolkl:
Tried to address the comments. Many thanks Benedikt! Also moved the filename property back to the config file, since that is not strictly related to pileup. Also at some point, the beamspot shape will have to be specified here.

Generation/python/FCCPileupScenarios.py
```diff
@@ -0,0 +1,32 @@
+
+from GaudiKernel import SystemOfUnits as units
+
+
+class FCCPileupScenarioBase(object):
+    """Beam Parameters Common to all FCC scenarios. Taken from
+    Benedikt, Michael, Daniel Schulte, and Frank Zimmermann:
```
hegner: Michael Benedikt, Daniel Schulte, and Frank Zimmermann

Generation/python/FCCPileupScenarios.py
```diff
@@ -0,0 +1,32 @@
...
+          "xVertexMin": -0.5 * units.mm,
+          "xVertexMax": 0.5 * units.mm,
+          "yVertexMin": -0.5 * units.mm,
+          "yVertexMax": 0.5 * units.mm,
+          "zVertexMin": -35 * units.mm,
+          "zVertexMax": 35 * units.mm
+        }
+        self.PileupToolConfig = {}
+
+class FCCPhase1Pileup(FCCPileupScenarioBase):
```
hegner: Why two subclasses and not just two instances of a pile up scenario class?

vvolkl: Makes a lot more sense -- I went one step further and used simple dictionaries instead (for now).

hegner:
retest this please

PR #88, opened 2016-03-11
replace v00 cards

selvaggi:

PR #89, opened 2016-03-14
Fixing Delphes configuration

jlingema:
I reverted Clement's bugfix since that seems to break the Delphes package with the current card.

@clementhelsens I guess this fix is needed with the new card? In that case maybe we should add the change to PR #88 .

Cheers,
Joschka

PS: [Nightlies](https://phsft-jenkins.cern.ch/job/FCCSW/BUILDTYPE=Debug,COMPILER=gcc49,LABEL=fcc-slc6/65/) failing due to the change.


clementhelsens:
Thanks!
yes, right, I thought I'd merge this PR at the same time. Thanks for asking Michele to add the bugfix in there.

PR #92, opened 2016-03-16
Treatment in HepMC reader for reaching end of file

jlingema:
- Changing some of the interfaces in the generation package to pass the HepMC event by reference instead of pointer
- Fix: The PoissonPileUp tool was returning the same number over and over, now it is random
- Fix: Memory leak in the HepMC file reader
- Change interface of merge tool: Explicitly take the main event that is merged with PU as a separate argument.

**Note** the reader tool will now return an error when the end of the file is reached before being done with the number of events specified. That also means that currently an error is produced when `EvtMax = -1`.


Generation/src/components/HepMCFileReaderTool.cpp
```diff
@@ -40,13 +40,20 @@ StatusCode HepMCFileReader::open(const std::string& filename) {
...
+    if (m_file->rdstate() == std::ios::eofbit) {
+        error() << "Error reading HepMC file" << endmsg;
+        return StatusCode::FAILURE;
+    }
+    // Reached the end of the file: Stop the run
+    IEventProcessor* eventProcessor;
+    service("ApplicationMgr",eventProcessor);
+    eventProcessor->stopRun();
+    error() << "Premature end of file: Please set the number of events according to hepMC file." << endmsg;
+    return Error("Reached end of file before finished processing");
```
jlingema: Here, we could not return the error but only do the `eventProcessor->stopRun()` to stop after the current event. However, I didn't like that because it could mean that we keep the last event where some of the pile-up events can't be read from the file.

hegner:
@jlingema - can you tackle the conflicts with the master? Thanks!

hegner:
retest this please

PR #93, opened 2016-03-24
Move some of the  markdown doc of examples to doxygen.

vvolkl:
Hi everyone,

I had a look at doxygen and noticed that it is possible to move a lot of the information of @zaborowska's nice [overview](https://github.com/HEP-FCC/FCCSW/blob/master/Examples/doc/ExamplesTable.md) of current example job options in the files themselves. We lose a bit on layout and style but gain on maintainability. (A preview is here: `file:///afs/cern.ch/user/v/vavolkl/public/FCCSW/doxygen/html/modules.html`) ) I also did not like how doxygen lists a separate namespace for each single python file, so I added __init__ files in the options directories -- this may be a bit confusing since there is not really an options module, but doxygen becomes a bit cleaner (compare the [old version](http://fccsw.web.cern.ch/fccsw/FCCSW/namespaces.html) to `file:///afs/cern.ch/user/v/vavolkl/public/FCCSW/doxygen/html/namespaces.html` ) . This is a question of style and opinion for sure, so let me know what you think!


jlingema:
Thanks, Valentin. I agree that this eases maintainability and the readability is still good enough, I think.. I'll merge this after tests succeed.

jlingema:
test this please

PR #94, opened 2016-03-29
Refactoring Delphes module

jlingema:
As discussed with @drasal, I did some changes to the Delphes module in FCCSW: Most notably, I removed the HepMC reader functionality that is already provided by the Generation package and moved the EDM conversion into dedicated tools to allow the user to select only the output that is needed. 

The results were tested against the output of the old version of the code, when using the same rnd-number seeds in Delphes and Pythia. For reference, I used [this](https://github.com/jlingema/fcc-physics/blob/delphes/example/compare_delphes_trees.py) to compare the EDM output files.


Sim/SimDelphesInterface/options/PythiaDelphes_config.py
```diff
@@ -55,25 +57,58 @@
...
+jetSaveTool.DataOutputs.jetConstituents.Path = "jetParts"
+jetSaveTool.DataOutputs.jetConstituentAssociations.Path = "jetsToParts"
+jetSaveTool.DataOutputs.jetFlavorTags.Path = "jetsFlavor"
+jetSaveTool.DataOutputs.jetFlavorTagAssociations.Path = "jetsToFlavor"
+jetSaveTool.DataOutputs.bTags.Path = "bTags"
+jetSaveTool.DataOutputs.cTags.Path = "cTags"
+jetSaveTool.DataOutputs.tauTags.Path = "tauTags"
+jetSaveTool.DataOutputs.jetBTagAssociations.Path = "jetsToBTags"
+jetSaveTool.DataOutputs.jetCTagAssociations.Path = "jetsToCTags"
+jetSaveTool.DataOutputs.jetTauTagAssociations.Path = "jetsToTauTags"
```
hegner: Any chance of rationalizing that in a clever way? Looks repetitive.

hegner:
If you do not want to have any other work in the pipeline for that PR I am happy to merge.

Sim/SimDelphesInterface/src/DelphesSimulation.cpp
```diff
@@ -193,24 +101,21 @@ StatusCode DelphesSimulation::initialize() {
...
-  m_chargedOutArray  = nullptr;
-  m_neutralOutArray  = nullptr;
-  m_photonOutArray   = nullptr;
-  m_jetOutArray      = nullptr;
-  m_metOutArray      = nullptr;
-  m_shtOutArray      = nullptr;
 
   m_eventCounter     = 0;
 
   if (m_outRootFile!=nullptr) m_treeWriter->Clear();
```
jlingema: Also should apply the same bugfix as in #97 here, when resolving conflicts.

jlingema:
retest this please

PR #100, opened 2016-04-29
Cleaning up toolchain.cmake

jlingema:
Removing check for LCG_generators as it was not used anymore, removing obsolete code.


jlingema:
@clementhelsens: This PR is needed to use the cvmfs setup.

clementhelsens:
Thanks, I'll merge as we are testing cvmfs on the grid

PR #101, opened 2016-05-09
New Implementation of merge tool, keeping all HepMC info on pileup, …

vvolkl:
... instead of just final state particles. This should help overcome the problem of G4PrimaryVertices overflowing the stack, when one is created for each particles. There is a conflict with #92, which updates the interface for the merge tool, depending which is merged first, one should be adapted.


jlingema:
test this please

Generation/data/Pythia_minBias.cmd
```diff
@@ -0,0 +1,65 @@
+!
+! File: Pythia_standard.cmd
```
jlingema: I guess this should read `minBias` rather than `standard`?

jlingema: Maybe we should also add CM energy + colliding particles to the file-title? E.g. `minbias_pp_10TeV`

Generation/src/components/HepMCFullMerge.cpp
```diff
@@ -0,0 +1,59 @@
...
+  if (!sc.isSuccess()) return sc;
+  return sc;
+}
+
+HepMC::GenEvent* HepMCFullMerge::merge(const std::vector<HepMC::GenEvent>& eventVector) {
+  // initial event is assumed to be signal event -- copied in full
+  HepMC::GenEvent* pileEvent = new HepMC::GenEvent(eventVector[0]);
+  for (auto it=eventVector.cbegin()+1, end = eventVector.cend(); it != end; ++it) {
+    // keep track of which vertex in full event corresponds to which vertex in merged event
+    std::map<const HepMC::GenVertex*, HepMC::GenVertex*> inputToMergedVertexMap;
```
jlingema: Can you change this to an `unordered_map`?

Generation/src/components/HepMCFullMerge.h
```diff
@@ -0,0 +1,36 @@
...
+
+#include "GaudiAlg/GaudiTool.h"
+#include "GaudiKernel/RndmGenerators.h" 
+
+#include "HepMC/IO_GenEvent.h"
+#include "HepMC/GenEvent.h"
+
+
+/** 
+ * Implementation partly due to pilemc https://pilemc.hepforge.org/
```
jlingema: Could you also add a line what the difference is w.r.t. the `HepMCSimpleMerge` tool?

Generation/src/components/HepMCSimpleMerge.cpp
```diff
@@ -31,20 +31,27 @@ HepMC::GenEvent* HepMCSimpleMerge::merge(const std::vector<HepMC::GenEvent>& eve
   HepMC::GenEvent* pileEvent = new HepMC::GenEvent(eventVector[0]);
   // iterate over vertices and add them to pileEvent
   for (auto it=eventVector.cbegin()+1, end = eventVector.cend(); it != end; ++it) {
+    std::map<const HepMC::GenVertex*, HepMC::GenVertex*> inputToMergedVertexMap;
```
jlingema: This could also be an `unordered_map`..

Generation/data/Pythia_minBias.cmd
vvolkl: Sure! Also for the signal file?

Generation/src/components/PythiaInterface.cpp
```diff
@@ -32,17 +35,25 @@ StatusCode PythiaInterface::initialize() {
...
+
+  // Read Pythia configuration files
+  m_pythiaSignal->readFile( m_parfile.c_str() );
+  // do not bother with pileup configuration if no pileup
+  if (0. < m_pileUpTool->getMeanPileUp()) {
+    if (m_pileUpTool->getFilename().empty()) {
+     return Error ( "Define Pythia8 configuration file for pileup in pileuptool (*.cmd)!" );  
+    }
+    m_pythiaSignal->readFile(m_pileUpTool->getFilename().c_str());
+  }
```
jlingema: this looks to me like it should be part of the pile-up tool initialisation? I guess the problem is that we have only one Pythia8 instance? In any case, we should re-visit this as it assumes that pile-up is always generated with Pythia (but it might be read from HepMC file, e.g.)

Generation/src/components/PythiaInterface.cpp
```diff
@@ -32,17 +35,25 @@ StatusCode PythiaInterface::initialize() {
   if (System::getEnv("PYTHIA8_XML") != "UNKNOWN") xmlpath = System::getEnv("PYTHIA8_XML") ;
 
   // Initialize pythia
-  m_pythia = std::unique_ptr<Pythia8::Pythia>(new Pythia8::Pythia(xmlpath));
-
-  // Read Pythia configuration file
-  m_pythia->readFile( m_parfile.c_str() );
+  m_pythiaSignal = std::unique_ptr<Pythia8::Pythia>(new Pythia8::Pythia(xmlpath));
+  m_pythiaPileup = std::unique_ptr<Pythia8::Pythia>(new Pythia8::Pythia(xmlpath));
```
jlingema: I guess we only want one :-)

Generation/src/components/PythiaInterface.cpp
jlingema: Sorry, I mis-read the above pointer declaration. Now I see that we actually do have 2 pythia instances. 

So what we should have is a specialised PileUp tool for generating PU with Pythia. That tool then owns the corresponding Pythia instance and then we can move this there as well.

Generation/src/components/PythiaInterface.cpp
```diff
@@ -73,30 +85,36 @@ StatusCode PythiaInterface::execute() {
   // Print debug: Pythia event info
   if (msgLevel() <= MSG::DEBUG) {
 
-    for (int i = 0; i < m_pythia->event.size(); ++i){
+    for (int i = 0; i < m_pythiaSignal->event.size(); ++i){
 
-      //if (m_pythia->event[i].isFinal() && m_pythia->event[i].isCharged())
+      //if (m_pythiaSignal->event[i].isFinal() && m_pythiaSignal->event[i].isCharged())
```
jlingema: Please remove commented out code.

hegner:
restest this please

hegner:
retest this please

jlingema:
ok to test

PR #102, opened 2016-05-11
Merging conflicts #82

clementhelsens:
Merging conflicts in HEP-FCC/FCCSW/#82

PR #82, opened 2016-03-04
adding particle generators

Ada15764:
Adding a set of classes that act as G4 particle generators (new files). no side effect is expected.

PR #103, opened 2016-05-19
Updated particle gun and magnetic field. Adding documentation for DD4hep

jlingema:
Some code clean-up in the simulation package and of specifically of #82, this supersedes #102. Improving documentation of magnetic field tools.


jlingema:
retest this please

jlingema:
Hi @dellacqu,

Apparently, the "NystromRK4" stepper sometimes fails with a division by zero (only caught when using Geant 4 compiled with the debug flags set and not occurring every time, see crashes from [your original PR](https://phsft-jenkins.cern.ch/job/FCCSW-pullrequests/95/BUILDTYPE=Debug,COMPILER=gcc49,LABEL=fcc-slc6/console) and [this PR](https://phsft-jenkins.cern.ch/job/FCCSW-pullrequests/132/BUILDTYPE=Debug,COMPILER=gcc49,LABEL=fcc-slc6/console)). We're setting the default to the "G4ClassicalRK4" stepper for now. Could you have a look if the defaults for the various stepper parameters make sense for the NystromRK4 stepper?

Thanks,
Joschka

jlingema:
retest this please

jlingema:
rettest this please

Sim/SimG4Components/src/G4ConstantMagneticFieldTool.cpp
```diff
@@ -1,137 +1,117 @@
...
+StatusCode G4ConstantMagneticFieldTool::initialize() {
+  StatusCode sc = GaudiTool::initialize();
+  if (sc.isFailure()) return sc;
 
-G4ConstantMagneticFieldTool::~G4ConstantMagneticFieldTool()
-{
-}
+  if (m_fieldOn) {
+    m_field =
+        new sim::G4ConstantField(m_fieldComponentX, m_fieldComponentY, m_fieldComponentZ, m_fieldRadMax, m_fieldZMax);
```
hegner: this line and the line where the field is passed to the fieldManager should be closer together and have a comment, who owns the field in the end.

jlingema: OK, I added a comment (ownership stays with the tool, the manager only has an observing pointer).

jlingema:
Clement has discussed the stepper issue with Andrea and the outcome was that he's fine with keeping the G4ClassicalRK4 for now. The reason for using the NystromRK4 stepper was that it is faster. There should be a follow-up why it crashes in case we want to switch in the future.

So from my point of view it is fine to merge this now, unless there are any other comments.

hegner:
@jlingema - all fine from my side.

PR #104, opened 2016-05-19
Change init to pick up stack initialisation from afs

jlingema:
The initialisation script now picks up the one we install in afs.


jlingema:
retest this please

PR #105, opened 2016-05-20
Extended format checker

jlingema:
The format checker now can also be used to check whole directories and to apply the changes in-place. I also updated the documentation accordingly.


jlingema:
retest this please

PR #110, opened 2016-06-06
Fixing input declaration in the Sim package

jlingema:
After the recent change to introduce the particle gun in the sim package, input was falsely still declared for the algorithm, though it is actually consumed through a tool. This PR moves the declaration to the tool to make it configurable.


hegner:
@jlingema - I would have expected somehow a change to a config file as well...

jlingema:
@hegner - Of course you are absolutely right.. I jumped the gun with opening, will add the changes to the configs asap.

PR #114, opened 2016-07-07
Sensitive detectors

zaborowska:
1. Implementations of SD were changed/added. `SimpleTrackerSD` and `SimpleCalorimeterSD` are the default SD types to use, they create hit for each energy deposit.
Other two: `AggregateCalorimeterSD` - during simulation accumulates energy deposits in cells; `MiddleStepTrackerSD` - like `SimpleTrackerSD`, but uses different position to compute segmentation cell ID (middle between G4PreStepPoind and G4PostStepPoint instead of G4PreStepPoint)
2. Documentations: [Detector](http://github.com/zaborowska/FCCSW/tree/testCellId/Detector/doc/DD4hepInFCCSW.md) & [Sim](http://github.com/zaborowska/FCCSW/tree/testCellId/Sim/doc/Geant4fullsim.md) were updated
3. cellID is now computed correctly for the detector that is not using DD4hep segmentation.
4. DetFCChhTrackerParametric has been removed as it only defined envelope for the tracker (envelopes are now available in DetCommon)
5. Few tests are created to check if cell ID is computed as expected in GDML, DD4hep with a simple box + XYZ segmentation and in DD4hep + box filled with NxNxN cells.
That includes the whole set for the detector constructed from GDML (SD, hit), as well as dummy physics list that kills particles at their origin.

PR #115, opened 2016-07-11
New implementation of DetFCChhTrackerSimple

vvolkl:
DD4hep description of a simplistic tracker. The dimensions correspond roughly to the tklayout baseline <http://fcc-tklayout.web.cern.ch/fcc-tklayout/FCChh_Delphes/info.html> but the xml is meant to be easily configurable and extended rather than be very exact. The dimensions are given relative to the definitions in `FCCh_DectDimensions.xml`, so the tracker will scale a bit when changing those values (but not in a smart way, module dimensions stay the same, and eta-coverage will change).


hegner:
test this please

Sim/SimG4Components/tests/options/geant_fastsim_simple.py
```diff
@@ -19,7 +19,8 @@
 # DD4hep geometry service
 # Parses the given xml file
 from Configurables import GeoSvc
-geoservice = GeoSvc("GeoSvc", detectors=['file:../../../Detector/DetFCChhTrackerParametric/compact/ParametricSimTrackerStandalone.xml'])
+geoservice = GeoSvc("GeoSvc", detectors=['file:Detector/DetFCChhBaseline1/compact/FCChh_DectEmptyMaster.xml',
+  'file:Detector/DetFCChhTrackerSimple/compact/FCChh_GenericTracker.xml'],
```
jlingema: Why move away from the parametric (empty) tracker for fast simulation?

vvolkl: oh, one replace too many. I will revert this one.

jlingema:
retest this please

jlingema:
retest this please

jlingema:
retest this please

jlingema:
retest this please

Detector/DetFCChhTrackerSimple/src/GenericBarrel_Geo.cpp
```diff
@@ -0,0 +1,161 @@
+
+#include "DetCommon/DetUtils.h"
+
+#include "DD4hep/DetFactoryHelper.h"
+
+using DD4hep::Geometry::Volume;
+using DD4hep::Geometry::DetElement;
+using DD4hep::XML::Dimension;
+using DD4hep::Geometry::PlacedVolume;
```
jlingema: since you always use the full names below, you can delete these...

Detector/DetFCChhTrackerSimple/src/GenericBarrel_Geo.cpp
```diff
@@ -0,0 +1,161 @@
...
+    DD4hep::XML::Component xLayer = static_cast<DD4hep::XML::Component>(xLayerColl);
+
+    DD4hep::XML::Component xModuleComponents = xmlElement.child("module_components");
+    DD4hep::XML::Component xModule =
+        det::getNodeByStrAttr(xmlElement, "module", "name", xLayer.attr<std::string>("module"));
+
+    // optional parameters
+    double stereo_offset = det::getDoubleWithDefault(xLayer, "stereo_offset", 0.0);
+    double module_twist_angle = det::getDoubleWithDefault(xLayer, "module_twist_angle", 0.1 * M_PI);
+    double stereo_module_overlap = det::getDoubleWithDefault(xLayer, "stereo_module_overlap", 0.0);
```
jlingema: Can you add comments in the xml that you have these hardcoded defaults if they are not specified in the xml?

Detector/DetFCChhTrackerSimple/src/GenericBarrel_Geo.cpp
```diff
@@ -0,0 +1,161 @@
...
+    double layerThickness;
+    unsigned int numRepeat;
+    double layer_rmin;
+    // handle repeat tag in xml
+    if (xLayer.hasAttr("repeat")) {
+      // "repeat" layers  equidistant between rmin and rmax
+      numRepeat = xLayer.repeat();
+      layerThickness = (xLayer.rmax() - xLayer.rmin()) / numRepeat;
+      layer_rmin = xLayer.rmin();
+    } else {
```
jlingema: I still think we should always specify the repeat attribute and get rid of the default, here.

Detector/DetFCChhTrackerSimple/src/GenericBarrel_Geo.cpp
```diff
@@ -0,0 +1,161 @@
...
+      r = layer_rmin + repeatIndex * layerThickness;
+      // definition of layer volumes
+      DD4hep::Geometry::Tube layerShape(r, r + layerThickness, xLayer.dz());
+      std::string layerName = "layer" + std::to_string(idxLay);
+      Volume layerVolume(layerName, layerShape, lcdd.material("Silicon"));
+      layerVolume.setVisAttributes(lcdd.invisible());
+      PlacedVolume placedLayerVolume = topVolume.placeVolume(layerVolume);
+      placedLayerVolume.addPhysVolID("layer", idxLay);
+      // approximation of tklayout values
+      nPhi = static_cast<unsigned int>(2 * M_PI * r / (0.9 * 2 * xModule.width())) + 1;
```
jlingema: Can you motivate these numbers in a comment (specifically the factor 0.9 and the +1)

Detector/DetFCChhTrackerSimple/src/GenericEndcap_Geo.cpp
```diff
@@ -0,0 +1,133 @@
...
+  // definition of top volume
+  std::string detName = xmlDet.nameStr();
+  DetElement GenericTrackerEndcapWorld(detName, xmlDet.id());
+
+  // envelope volume with the max dimensions of tracker for visualization etc.
+  // contains both endcaps, in forward and in backwards direction
+
+  DD4hep::Geometry::Tube posnegEnvelopeShape_plus(dimensions.rmin(), dimensions.rmax(), (dimensions.z2()));
+  DD4hep::Geometry::Box posnegEnvelopeShape_minus(
+      dimensions.rmax() * 1.001, dimensions.rmax() * 1.001, dimensions.z1());
```
jlingema: Why do we have factors `1.001` here but not for the positive side?

Detector/DetFCChhTrackerSimple/src/GenericEndcap_Geo.cpp
```diff
@@ -0,0 +1,133 @@
...
+  // loop over 'layer' nodes in xml
+  unsigned int idxLay = 0;
+  for (DD4hep::XML::Collection_t xLayerColl(xmlElement, _U(layers)); nullptr != xLayerColl; ++xLayerColl) {
+    DD4hep::XML::Component xLayer = static_cast<DD4hep::XML::Component>(xLayerColl);
+
+    // TODO: define modules
+    // xml_comp_t xModule = det::getNodeByStrAttr(xmlElement, "module", "name", xLayer.attr<std::string>("module"));
+
+    // create petals
+    unsigned int nPhi = static_cast<unsigned int>(det::getDoubleWithDefault(xLayer, "nPhi", 16));
```
jlingema: Please document these defaults in the xml.

Detector/DetFCChhTrackerSimple/src/GenericEndcap_Geo.cpp
```diff
@@ -0,0 +1,133 @@
...
+        lcdd.material("Silicon"));
+    petalVolume.setVisAttributes(lcdd, xLayer.visStr());
+    // TODO: set modules sensitive instead of petals as soon as there are modules
+    petalVolume.setSensitiveDetector(sensDet);
+
+    // handle repeat tag in xml
+    double layerThickness;
+    unsigned int numLayers;
+    double current_z;
+    if (xLayer.hasAttr("repeat")) {
```
jlingema: same comment as for barrel: I think this should be a required attribute

jlingema:
Thanks a lot @vvolkl! I saw quite a few TODO / @todo in the code. Maybe you could create a JIRA ticket and keep track there, rather than in the code? 

Are there some you'd like to resolve before we merge (e.g. the half-length TODO from the endcap did sound like one that might cause confusion and could quickly solved)?

vvolkl:
Thanks for the comments! I think I can fix everything now, except for adding module structures to the endcaps, which still requires some thought and can be done together with adding extensions for the tracking geometry. I can create a JIRA ticket for this.

Detector/DetFCChhTrackerSimple/src/GenericBarrel_Geo.cpp
vvolkl: But I am using the `using` statements, e.g. https://github.com/HEP-FCC/FCCSW/pull/115/files#diff-6739f93c51529d9fd488b8cd20bc5214R30

vvolkl:
Note that currently, hits will only be created for *one* of the detectors with `SimpleTrackerSD`. This will be fixed by  #114.

Detector/DetFCChhTrackerSimple/src/GenericEndcap_Geo.cpp
vvolkl: The variable names were a bit misleading, this is actually the envolope for both sides, but I subtract the volume where I assume the Barrel, so there are no envelope-overlaps. I changed the variable names and added a comment.

jlingema:
retest this please.

Detector/DetCommon/DetCommon/DetUtils.h
```diff
@@ -0,0 +1,12 @@
+#ifndef DETECTOR_DETUTILS_H
+#define DETECTOR_DETUTILS_H
+
+#include "DD4hep/DetFactoryHelper.h"
+
+namespace det {
+DD4hep::XML::Component getNodeByStrAttr(const DD4hep::XML::Handle_t& mother, const std::string& nodeName, const std::string& attrName,
+                                        const std::string& attrValue);
+
+double getDoubleWithDefault(const DD4hep::XML::Component& node, const std::string& attrName, const double& defaultValue);
```
hegner: that method name is meaningless. what does it actually do?

Detector/DetCommon/src/DetUtils.cpp
```diff
@@ -0,0 +1,26 @@
...
+  for (DD4hep::XML::Collection_t xCompColl(mother, nodeName.c_str()); nullptr != xCompColl; ++xCompColl) {
+    if (xCompColl.attr<std::string>(attrName.c_str()) == attrValue) {
+      return static_cast<DD4hep::XML::Component>(xCompColl);
+    }
+  }
+  // in case there was no xml daughter with matching name
+  return DD4hep::XML::Component(nullptr);
+}
+
+/// try to get attribute with double value, return defaultValue if attribute not found
```
hegner: such a line may be better in the header.

Detector/DetFCChhTrackerSimple/src/GenericBarrel_Geo.cpp
```diff
@@ -0,0 +1,154 @@
...
+
+  // counts all layers - incremented in the inner loop over repeat - tags
+  unsigned int idxLay = 0;
+  // loop over 'layer' nodes in xml
+  for (DD4hep::XML::Collection_t xLayerColl(xmlElement, _U(layers)); nullptr != xLayerColl; ++xLayerColl) {
+    DD4hep::XML::Component xLayer = static_cast<DD4hep::XML::Component>(xLayerColl);
+
+    DD4hep::XML::Component xModuleComponents = xmlElement.child("module_components");
+    DD4hep::XML::Component xModule =
+        det::getNodeByStrAttr(xmlElement, "module", "name", xLayer.attr<std::string>("module"));
```
hegner: isn't the det:: unnecessary?

Detector/DetFCChhTrackerSimple/src/GenericBarrel_Geo.cpp
```diff
@@ -0,0 +1,154 @@
...
+      DD4hep::Geometry::Position offset(0, -0.5 * totalModuleComponentThickness + integratedModuleComponentThickness,
+                                        0);
+      integratedModuleComponentThickness += xComp.thickness();
+      PlacedVolume placedModuleComponentVolume = moduleVolume.placeVolume(moduleComponentVector[idxSubMod], offset);
+      placedModuleComponentVolume.addPhysVolID("module_component", idxSubMod);
+    }
+
+    double layerThickness;
+    unsigned int numRepeat;
+    double layer_rmin;
```
hegner: (maybe something for the coding conventions) always use initialized variables.

Detector/DetFCChhTrackerSimple/src/GenericBarrel_Geo.cpp
```diff
@@ -0,0 +1,154 @@
...
+    // handle repeat attribute in xml
+    // "repeat" layers  equidistant between rmin and rmax
+    numRepeat = xLayer.repeat();
+    layerThickness = (xLayer.rmax() - xLayer.rmin()) / numRepeat;
+    layer_rmin = xLayer.rmin();
+    unsigned int nPhi = 0;
+    double r = 0;
+    double phi = 0;
+    // loop over repeated layers defined by one layer tag
+    for (unsigned int repeatIndex = 0; repeatIndex < numRepeat; ++repeatIndex, ++idxLay) {
```
hegner: can you explain why idxLay doesn't need to be set to zero before?

Detector/DetFCChhTrackerSimple/src/GenericEndcap_Geo.cpp
```diff
@@ -0,0 +1,125 @@
...
+#include "DD4hep/DetFactoryHelper.h"
+
+using DD4hep::Geometry::Volume;
+using DD4hep::Geometry::DetElement;
+using DD4hep::XML::Dimension;
+using DD4hep::Geometry::PlacedVolume;
+
+namespace det {
+/**
+Factory for a configurable, generic tracker endcap.
```
hegner: Do you somewhere describe what the general layout of that endcap is?

Detector/DetFCChhTrackerSimple/src/GenericBarrel_Geo.cpp
vvolkl: It counts all layers in the detector, which may have seval `layers` elements in the xml, which again specifies several (`repeat`-many) layers. The name could probably be more descriptive.

Detector/DetCommon/DetCommon/DetUtils.h
vvolkl: I will move the comment from the body here! I don't know a better name though - the xml method is called getAttr and I call it several times and use a default when it is not available.

Detector/DetFCChhTrackerSimple/src/GenericBarrel_Geo.cpp
hegner: In that case please move that ++idxLay into the body of the loop. That makes the semantics more clear.

Detector/DetCommon/DetCommon/DetUtils.h
hegner: what about "getAttrValueWithDefault" or "getAttrValueWithFallback"?

Detector/DetFCChhTrackerSimple/src/GenericEndcap_Geo.cpp
vvolkl: I wrote a short comment in the xml: https://github.com/vvolkl/FCCSW/blob/generic_tracker/Detector/DetFCChhTrackerSimple/compact/Tracker.xml#L46

Detector/DetCommon/DetCommon/DetUtils.h
vvolkl: Sounds good! I am going with the last one.

vvolkl:
Ok, counter seems better.

hegner:
retest this please

hegner:
what concerns me it is ready :-) @jlingema - any further comments?

jlingema:
Sorry for the delay... For some reason I did not receive an email notification for your comment, @hegner. No further comments, so I'll  go ahead and merge.

PR #116, opened 2016-07-11
Fix example xml to work with updated type SimpleBox.

vvolkl:
Was broken. Not sure if needed, but might be useful for a quick test.


jlingema:
This isn't used in any configuration? Could you add one to have it tested?

vvolkl:
It is not used in any active test (But it turned up in an obsolete file, which I delete with the last commit). I can add a simple one!

jlingema:
test this please

PR #117, opened 2016-07-12
Adding reader algorithm

jlingema:
Some changes in the DataHandle were necessary, some of which are temporary until we migrate to Gaudi v27r1 or later (see remarks in the code).


Examples/src/CheckCollection.cpp
```diff
@@ -0,0 +1,51 @@
+#include "GaudiAlg/GaudiAlgorithm.h"
+
+#include "FWCore/DataHandle.h"
+
+#include "datamodel/MCParticleCollection.h"
+#include "datamodel/TrackHitCollection.h"
+#include "datamodel/GenVertex.h"
+
+
+class CheckCollection : public GaudiAlgorithm {
```
hegner: CheckCollection sounds too generic to me. Can you use a name like FCCTestConsumer or something like that?

FWCore/FWCore/DataHandle.h
```diff
@@ -61,9 +80,40 @@ DataHandle<T>::DataHandle() {}
...
+    if (LIKELY(m_isGoodType)) {
+      DataObjectHandle<DataWrapper<T> >::setRead();
+      std::cout << "returning static cast" << std::endl;
+      return static_cast<DataWrapper<T>*>(dataObjectp)->getData();
+    } else {
+      // The reader does not know the specific type of the collection. So we need a reinterpret_cast if the handle was
+      // created by the reader.
+      DataWrapper<podio::CollectionBase>* tmp = static_cast<DataWrapper<podio::CollectionBase>*>(dataObjectp);
+      DataObjectHandle<DataWrapper<T> >::setRead();
+      return reinterpret_cast<const T*>(tmp->collectionBase());
```
hegner: I am sure we can introduce some better type safety here. Let's discuss in person. Shouldn't be a blocker for this pull request though.

FWCore/components/PodioInput.h
```diff
@@ -0,0 +1,58 @@
...
+
+// forward declarations
+// from FWCore:
+class PodioDataSvc;
+
+
+/** @class PodioInput FWCore/components/PodioInput.h PodioInput.h
+ *
+ *  Class that allows to read ROOT files written with PodioOutput
+ *  TODO: Automatically load needed libraries (needs changes to scheduler), check nEvents and stop run if EOF is reached
```
hegner: make a JIRA ticket out of it ;-)

FWCore/components/PodioInput.h
```diff
@@ -0,0 +1,58 @@
...
+  /// ROOT file name the input is read from. Set by option filename
+  std::string m_filename;
+  /// Name of collections to read. Set by option collections (this is temporary)
+  std::vector<std::string> m_collectionNames;
+  /// Collection IDs (retrieved with CollectionIDTable from ROOT file, using collection names)
+  std::vector<int> m_collectionIDs;
+  /// Data service: needed to register objects and get collection IDs. Just an observing pointer.
+  PodioDataSvc* m_podioDataSvc;
+  /// Counter of the event number
+  int m_eventNum;
```
hegner: why not put the counter into the service?
Once we go for multithreading that may be require anyhow.

FWCore/src/PodioDataSvc.cpp
```diff
@@ -38,6 +39,13 @@ StatusCode PodioDataSvc::clearStore()    {
   return StatusCode::SUCCESS ;
 }
 
+void PodioDataSvc::setCollectionIDs(podio::CollectionIDTable* collectionIds) {
+  if (m_collectionIDs != nullptr) {
```
hegner: that makes me wonder whether a smart pointer could be something...

jlingema: As discussed: I agree but some changes to PODIO interfaces are necessary to make that a neat solution.

FWCore/components/PodioInput.h
jlingema: As discussed, I'll do this in a follow-up PR that moves the ownership of file, reader and event-store to the DataSvc.

FWCore/FWCore/DataHandle.h
jlingema: I added `dynamic_casts` (done once) further up to check if it is safe to do the `static_cast` and `reinterpret_cast` here.

PR #120, opened 2016-07-27
Automated Testing for Detector Geometry Overlaps

vvolkl:
Also tries to fix the overlaps currently reported. There are still some overlaps with the beamtube and the magnets, but maybe their problem with the shape conversion should be fixed first?

### DIFF
--- a/Detector/DetFCChhHCalTile/src/HCalBarrel_geo.cpp
+++ b/Detector/DetFCChhHCalTile/src/HCalBarrel_geo.cpp
@@ -33,6 +33,7 @@ static DD4hep::Geometry::Ref_t createHCal (
   // Let's skip this for now...
   // Det::DetCylinderVolume* detVolume = new Det::DetCylinderVolume(status);
   // hCal.addExtension<Det::IDetExtension>(detVolume);
+
   // Make volume that envelopes the whole barrel; set material to air
   Dimension dimensions(xmlDet.dimensions());
   DD4hep::Geometry::Tube envelopeShape(dimensions.rmin(), dimensions.rmax(), dimensions.dz());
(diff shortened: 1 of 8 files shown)


jlingema:
test this please

jlingema:
Hi Valentin,

This is great. Could you add a line in the DD4hep documentation on how to apply the overlap checks?

Thanks,
Joschka

Detector/DetFCChhHCalTile/src/HCalBarrel_geo.cpp
```diff
@@ -105,10 +105,10 @@ static DD4hep::Geometry::Ref_t createHCal (
   double dy0 = moduleDimensions.dz();
   double dz0 = moduleDimensions.dr() / 2.;
 
-  double drWedge = (dimensions.rmax() - dimensions.rmin()) * 0.5;
+  double drWedge = cos(dphi / 2.) * (dimensions.rmax() - dimensions.rmin()) * 0.5;
```
jlingema: @clementhelsens and @faltovaj - Valentin changed the dimensions of the modules slightly. From the xml description we take the total depth of the calorimeter and create that as a cylindrical envelope. Then we create trapezoids, previously those were slightly extruding from the envelope. This change reduces the sizes slightly (see attachment). Can you comment if we should go with this or rather adjust the envelopes?

geo_change

faltovaj: Hi Joschka,
I agree with the proposal. I think it is better to keep everything in the current envelope.
I was looking at the changes proposed for the FrontPlate and EndPlate. Seems like FrontPlate has a width of 0 at the moment. Valentine, could you please fix that? After that fix, we would need to correct barrelRmin,drWedge, numModulesR etc. to take into account that the active calorimeter starts at dimensions.rmin() + dRhoFacePlate. Also endPlateShape should start from here in r. Right?
Thanks,
Jana

vvolkl: Hi Jana,
right! Thanks for spotting this, I misread the plate dimensions. Ok, I will try to implement your suggestion so the plates are contained by the envelope.
Cheers,
Valentin

jlingema:
retest this please

jlingema:
The failing test is due to a new file structure introduced in #114 (which is already put on afs to be able to test the PR). Will retest this once that one is merged.

jlingema:
ok to test

PR #124, opened 2016-08-17
Detector utilities

zaborowska:
1. New segmentation in phi-eta was introduced (PhiEtaGrid):
    - cell size in eta (`grid_size_eta`)
    - number of cells in phi (`phi_bins`), resulting in phi cell size of 2pi/phi_bins
2. DetUtils are extended with the functions to:
    - find the neighbours of a cell (can be in any dimension, not only segmentation, e.g. layer)
    - get the total number of volumes with a name containg certain string (thanks to @vvolkl)
    - get the number of segmentation cells within the volume
    Examples can be found for [XYZ segm.](https://github.com/zaborowska/FCCSW/blob/detectorUtils/Test/TestReconstruction/tests/options/testcellcountingXYZ.py), [RPhi segm.](https://github.com/zaborowska/FCCSW/blob/detectorUtils/Test/TestReconstruction/tests/options/testcellcountingRPhi.py) and [PhiEta segm.](https://github.com/zaborowska/FCCSW/blob/detectorUtils/Test/TestReconstruction/tests/options/testcellcountingPhiEta.py)
3. Algorithms for the cell size optimisation/tests. 
- Merging adjacent cells  (`MergeCells`) can be used if initial simulation was made using very fine granular segmentation and one wants to study the effect of larger cells. Currently no offset in the segmentation is supported, meaning that if field is 'signed', the merging can be done only for odd number of cells (to keep middle of the middle cell in 0).
Example for [merging x cells from 5mm to 15mm](https://github.com/zaborowska/FCCSW/blob/detectorUtils/Detector/DetComponents/tests/options/mergeCells.py)
- Redoing the segmentation (`RedoSegmentation`) can be used if the segmentation needs to be changed in a different manner than a simple merging of cells (e.g. change of segmentation type). It requires the deposit position to be saved in cluster position. This may be changed once we migrate to newer EDM. The new segmentation is taken from the xml file with the detector description (user needs to create new `<readout>` tag, with new segmentation)
Examples: for [XYZ](https://github.com/zaborowska/FCCSW/blob/detectorUtils/Detector/DetComponents/tests/options/redoSegmentationXYZ.py) and [PhiEta](https://github.com/zaborowska/FCCSW/blob/detectorUtils/Detector/DetComponents/tests/options/redoSegmentationPhiEta.py)

If one wants to use own segmentation (such as `GridPhiEta`), DD4hep requires:
`export LD_PRELOAD=build.$BINARY_TAG/lib/libDetSegmentation.so`

Two new test geometries are added (next to existing box with only segmentation and box with volumes as cells w/o segmentation):
 - [barrel with r, phi, z cells](https://github.com/zaborowska/FCCSW/blob/detectorUtils/Test/TestGeometry/data/Barrel_testCaloSD_rphiz.xml) - with slices in z and segmentation in r, phi 
 - [barrel with phi, eta cells](https://github.com/zaborowska/FCCSW/blob/detectorUtils/Test/TestGeometry/data/Barrel_testCaloSD_phieta.xml) - one volume with phi, eta segmentation


jlingema:
ok to test

Detector/DetCommon/DetCommon/DetUtils.h
```diff
@@ -1,17 +1,126 @@
...
+ */
+std::array<uint, 2> numberOfCells(uint64_t aVolumeId, const DD4hep::DDSegmentation::PolarGridRPhi& aSeg);
+
+/** Get the number of the volumes containing a given name.
+ *   For an example see: Test/TestReconstruction/tests/options/testcellcountingXYZ.py.
+ *   @param[in] aHighestVolume The top volume in the geometry.
+ *   @param[in] aMatchName Name (or its part) of the volume.
+ *   return Number of the volumes.
+ */
+unsigned int countPlacedVolumes(TGeoVolume* aHighestVolume, std::string aMatchName);
```
jlingema: I'm guessing we can pass both arguments as `const` and the string as ref?

Detector/DetCommon/src/DetUtils.cpp
```diff
@@ -11,10 +23,160 @@ DD4hep::XML::Component det::getNodeByStrAttr(const DD4hep::XML::Handle_t& mother
...
+}
+
+std::vector<std::pair<int,int>> bitfieldExtremes(DD4hep::DDSegmentation::BitField64& aDecoder,
+    const std::vector<std::string>& aFieldNames) {
+  std::vector<std::pair<int,int>> extremes;
+  int width = 0;
+  for(const auto& field: aFieldNames) {
+    width = aDecoder[field].width();
+    if( aDecoder[field].isSigned() ){
+      extremes.emplace_back(std::make_pair( ( 1LL << ( width - 1 ) ) - ( 1LL << width ),
```
jlingema: This is the same as -(1LL << (width - 1)), which is a bit easier to understand, I think...

Detector/DetCommon/src/DetUtils.cpp
```diff
@@ -11,10 +23,160 @@ DD4hep::XML::Component det::getNodeByStrAttr(const DD4hep::XML::Handle_t& mother
...
+          ( 1LL << ( width - 1 ) ) - 1));
+    } else {
+      extremes.emplace_back(std::make_pair( 0,
+         (0x0001 << width) - 1 ));
```
jlingema: Can you keep a consistent notation? Either is fine (`0x0001` or `1LL`).

jlingema: Although, thinking about this a bit more:
1) `1LL` is better because it explicitly notes the size
2) But do we actually expect 64 bit here? Since this is the individual fields, I assume it's possible, but unlikely?
3) If we work with `long long` the return type should reflect that

Detector/DetCommon/src/DetUtils.cpp
```diff
@@ -11,10 +23,160 @@ DD4hep::XML::Component det::getNodeByStrAttr(const DD4hep::XML::Handle_t& mother
...
+
+std::array<uint, 2> numberOfCells(uint64_t aVolumeId, const DD4hep::DDSegmentation::CartesianGridXY& aSeg) {
+  // // get half-widths
+  auto halfSizes = envelopeDimensions(aVolumeId);
+  // get segmentation cell widths
+  double xCellSize = aSeg.gridSizeX();
+  double yCellSize = aSeg.gridSizeY();
+  // calculate number of cells, the middle cell is centred at 0 (no offset)
+  uint cellsX = ceil((halfSizes.x()-xCellSize/2.)/xCellSize)*2+1;
+  uint cellsY = ceil((halfSizes.y()-yCellSize/2.)/yCellSize)*2+1;
```
jlingema: Why the `ceil` and `-yCellSize/2.` and `+1`? Shouldn't just the `ceil` without the other two do the trick?

zaborowska: I believe that wouldn't take into account that middle cell is centred at 0 (and that means we have odd number of cells):
e.g.
detector half size = 10, cell size = 2:
ceil((10-1)/2) *2 +1 = ceil(4.5) *2 +1 = 11
what I understand you suggest:
ceil(10/2) *2 = 10

PR #125, opened 2016-08-31
Necessary changes to run SAS

jlingema:


CMakeLists.txt
```diff
@@ -1,20 +1,19 @@
 CMAKE_MINIMUM_REQUIRED(VERSION 2.8.5)
+
+option(SAS "Whether to run static code analysis with SAS." OFF)
+if (SAS)
+    find_package(sas)
+    enable_sas(
+        FORMAT
+        COMPARISON_REPORT_DIR "${CMAKE_BINARY_DIR}/sas_report"
+        IGNORE_DIRS ${CMAKE_SOURCE_DIR}/build ${CMAKE_SOURCE_DIR}/cmake ${CMAKE_SOURCE_DIR}/doc
```
hegner: What about moving ignoring the build dir (expressed in terms of CMAKE_BINARY_DIR) into `enable_sas`?

jlingema: Actually already discussed that with David and the corresponding [PR](https://github.com/jlingema/SAS/pull/9/files#diff-2ff601796f9e6f919227d31176e88c0fR78) is waiting for a conflict resolve :-)

jlingema: However, I'd like to keep this in, because it also allows to ignore all other build directories (i.e. if you have both debug and release builds).

hegner: I understand. Maybe we should improve on the out-of-source build in the long run. Fine for merging if SAS already has that change in the pipeline.

PR #126, opened 2016-08-31
Moving fast sim config

zaborowska:
Fast simulation configuration has been moved from Run User Action to specific tools that create regions.
Region tools can be passed to `SimG4Svc`.
Currently there is an implementation of a tool that creates a region and attaches a fast sim model for tracker smearing.
The model configuration has been moved from smearing tools to region tool (thus avoiding the code duplication).


jlingema:
ok to test

PR #127, opened 2016-08-31
Small fixes

zaborowska:
1. Fixing test so that it actually saves the hits from calo (not only creates a tool to do it)
2. Adding checks to particle gun
3. Adding comment that using segmentation from outside DD4hep requires adding its lib to `LD_PRELOAD_PATH`


jlingema:
ok to test

PR #128, opened 2016-09-19
Fixing bugs in detector utilities

zaborowska:
Fixing a calculation of bins in eta (adding forgotten calculation of eta);
Updating GridPhiEta segmentation - using offset in eta to get a proper cellID.


jlingema:
ok to test

PR #132, opened 2016-10-31
Fix  new DD4hep runtime error in HCal construction 

vvolkl:
More recent versions of DD4hep give a runtime error when adding physVolIds that are not defined in the readout.


hegner:
test this please

PR #133, opened 2016-10-31
Commit script for running Pythia + Delphes 

selvaggi:


jlingema:
Hi Michele,

I did a little bit of digging and found that `gaudirun.py` supports customisation in a very generic way using the `--option` optional argument, e.g. to set the max number of events you can do:

```
./run gaudirun.py Sim/SimDelphesInterface/option/PythiaDelphes_config.py --option "AppManager.EvtMax=10"
```

The content of `option` can be any python code and is appended to the configuration. Since this depends on the naming in the configuration, I don't think it makes a lot of sense to write a wrapper in FCCSW but rather document this option so users can use it as is.

PR #135, opened 2016-11-01
New algorithm for merging volumes  (readout: volID in hits collection)

zaborowska:
The existing merge algorithm, `MergeCells` enables to merge the cells of the segmentation into twice/three times/... larger cells by merging 2/3/... adjacent cells, starting from the cell with ID=0.

There is also need to merge volumes (not segmentation in DD4hep's meaning). In this case user knows perfectly how many volumes exist in the geometry and how to combine the signal from them for the readout.
It means, in particular, that user specifies to combine the first 2 layers together, then next 3, next 5, etc. So instead of one number, the list is taken. Also, there is a check if its sum agrees with the total number of volumes.
It diverges a lot from MergeCells so a new alg was created.
If anyone has a better idea for a name (something with readout? id?), I'd be glad to hear.


jlingema:
ok to test

hegner:
As discussed, maybe talking about layers makes the intention clearer.

PR #138, opened 2016-11-02
Small corrections to Delphes FCCSW interface

selvaggi:


jlingema:
Hi Michele,
This also needs the new Delphes version, right? If so, I'll install in the pre-release area so we can test the PR.
Cheers,
Joschka

selvaggi:
Yes, that would be great, thanks.

selvaggi:
Added new features such as event weight and and gen particle skimming.
Skimming originally planned to be donned in Delphes, but it gives more flexibility to do it FCCSW.
The un-skimmed MC particle collection is not stored by default, but can be easily done by removing the relevant drop in the PythiaDelphes_config.py file.

jlingema:
OK, this now also depends on HEP-FCC/fcc-edm#30

jlingema:
ok to test

Sim/SimDelphesInterface/options/PythiaDelphes_config.py
```diff
@@ -38,29 +38,29 @@ def apply_paths(obj, names):
...
+  
 ############################################################
 #
 # User: Configure variables
 #
 ############################################################
 
 ## N-events
-nEvents=100
+nEvents=1000
```
jlingema: Can we change this back to `100` to make the tests for the PRs shorter?

selvaggi:
done

jlingema:
retest this please

PR #142, opened 2016-11-07
Fast simulation for calorimeters: GFlash

zaborowska:
Configuration of the GFlash parametrisation implemented in Geant4.
It includes:
- new sensitive detector that can process deposits from full sim but also from parametrisation (produced hits are almost identical: gflash hits do not have time info)
- new xml for ECal that uses that SD
- region tool that creates region & model for GFlash with settings: energy thresholds, parametrisation configuration:
   - parametrisation of the homogeneous calorimeter using original parameters from arXiv:hep-ex/0001020v1 (already in G4)
   - parametrisation of the sampling calorimeter using original parameters from arXiv:hep-ex/0001020v1 (already in G4)
   - ongoing work on implementation of the parametrisation that would use auto-generated params within FCCSW (to be presented in next PR)
- updated documentation

Examples are in:
[Examples/options/geant_fastsim.py](https://github.com/HEP-FCC/FCCSW/compare/master...zaborowska:gflashmodel?expand=1#diff-e7eb64d66fc804d00bc22203d93f382b) for sampling calo (our currently used LAr-Pb calo)
[Test/TestGeometry/tests/options/gflash_test_pbwo4.py](https://github.com/HEP-FCC/FCCSW/compare/master...zaborowska:gflashmodel?expand=1#diff-33866548c9066578e9fa35a0736fca65) for homogeneous calo


Sim/SimG4Fast/src/components/SimG4GflashSamplingCalo.cpp
```diff
@@ -0,0 +1,56 @@
...
+    return StatusCode::FAILURE;
+  }
+  G4NistManager* nist = G4NistManager::Instance();
+  for(const auto& mat: m_materials) {
+    if(nist->FindOrBuildMaterial(mat) == nullptr) {
+      error()<<"Material <"<<mat<<"> is not found by G4NistManager."<<endmsg;
+      return StatusCode::FAILURE;
+    }
+  }
+  if(m_thickness.size() != 2) {
```
jlingema: If the `2` is hardcoded here anyway, maybe make it two explicit properties?

PR #144, opened 2016-11-23
move FCCDelphesTutorial to doc

selvaggi:


jlingema:
Skipping the test since it's documentation only.

PR #146, opened 2016-11-24
[WIP] First version of FCC-ee detector description

nalipour:


jlingema:
Hi Nilou,

Can you remove the materials.xml and elements.xml and instead add any missing entries into the ones that are in `Detector/DetCommon/compact/`?

It would also be great if you could add a test that constructs the detector and checks for overlaps:
see the example  [here](https://github.com/HEP-FCC/FCCSW/blob/master/Examples/CMakeLists.txt#L71), adding another config like [printOverlap.py](https://github.com/HEP-FCC/FCCSW/blob/master/Examples/options/printOverlaps.py#L5) but changing the detector that is used. That way we can automatically test that the detector description works properly.

Cheers,
Joschka

nalipour:
Dear Joschka,

Thanks for your comments. I have updated the materials and elements file. I have also added the printOverlap_FCCee.py to check the overlaps. Please let me know if I am missing anything else.

Cheers,
Nilou

jlingema:
Hi Nilou,

Many thanks. Can you also add it as a test, like is done [here](https://github.com/HEP-FCC/FCCSW/blob/master/Examples/CMakeLists.txt#L71)? You just need to copy the whole `gaudi_add_test` and point it to the new python script as well as rename it (the first argument). That way we can do the automatic testing.

Cheers,
Joschka

jlingema:
Thanks. The following allows us to test automatically.

test this please

jlingema:
Superseded by #187

PR #151, opened 2016-12-05
Adding command line argument to modify Delphes card

clementhelsens:
Add delphes card as possible argument and add a generic pythia command for batch

### DIFF
--- a/FWCore/python/FWCore/joboptions.py
+++ b/FWCore/python/FWCore/joboptions.py
@@ -4,11 +4,13 @@ def parse_standard_job_options(scriptname=""):
     """
     Returns the parsed arguments, adding a parser with commonly needed opts:
     - args.nevents      -- number of events (int), specify with --nevents
+    - args.delphescard  -- the input delphes card (string), specify with --delphescard
     - args.inputfile    -- the input file (string), specify with --inputfile
-    - args.outputfile   -- the output file (string), specifu with --outputfile
+    - args.outputfile   -- the output file (string), specify with --outputfile
     """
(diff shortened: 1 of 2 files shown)


FWCore/python/FWCore/joboptions.py
```diff
@@ -4,11 +4,13 @@ def parse_standard_job_options(scriptname=""):
     """
     Returns the parsed arguments, adding a parser with commonly needed opts:
     - args.nevents      -- number of events (int), specify with --nevents
+    - args.delphescard  -- the input delphes card (string), specify with --delphescard
```
jlingema: The idea was to have "standard" options here and at the same time allow parsers for specific things in the options file. E.g. like [this](https://gist.github.com/jlingema/d170c552bd3caf71a3d713ccf04d5fd5)

clementhelsens:
Hello, 

I made the requested changes.

Sim/SimDelphesInterface/options/PythiaDelphes_config.py
```diff
@@ -40,6 +40,12 @@ def apply_paths(obj, names):
   for attr, name in names.iteritems():
     getattr(obj.DataOutputs, attr).Path = name
 
+
+import argparse
+parser = argparse.ArgumentParser()
+parser.add_argument('--delphescard', type=str, default='', help='specify an input delphes card')
```
hegner: if the default is an invalid value - and an empty string is - one should say default=None. I am nitpicking though.

clementhelsens:
done!

PR #153, opened 2016-12-05
Fix some overlaps in TKLayout Detector

andresailer:
Component placement was not properly done.
The offset needs to be the centre of the component, so we need to add half the thickness of the current component. The overlaps were leading some components to be completely covered by others (depending on thicknesses), and this hits were not properly recorded. Without overlaps we can also only set only the actual sensitive material to be sensitive.
(Geant4 /geometry/test/run showed the overlaps. I tested this via the copy of the files in lcgeo, did not compile or run this in fccsw)


vvolkl:
Thanks for the fix! I wonder why TGeo didn't report this -- we check for overlaps by  doing ( in principle) `gGeoManager->CheckOverlaps(); gGeoManager->PrintOverlaps();` I'll have to look into that.

jlingema:
Thanks Andre!

ok to test

vvolkl:
It seems there are still some overlaps that were not detected. I'm following up with the ROOT experts and will fix it in a future PR.

PR #166, opened 2017-02-09
set default pythia8 generation parameters

selvaggi:


clementhelsens:
Thanks for the bugfix!

PR #167, opened 2017-02-09
Remove fixed seed for LHE generation

selvaggi:
also changed max errors and default number of events to generate

### DIFF
--- a/Generation/data/Pythia_LHEinput.cmd
+++ b/Generation/data/Pythia_LHEinput.cmd
@@ -7,10 +7,8 @@
 ! Adjusted from Pythia example: main42.cmnd
 
 ! 1) Settings that will be used in a main program.
-Main:numberOfEvents = 1            ! number of events to generate
-Main:timesAllowErrors = 3          ! abort run after this many flawed events
-Random:seed = 1234                ! initialize random generator with a seed
-
+Main:numberOfEvents = 100          ! number of events to generate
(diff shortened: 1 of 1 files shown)


Generation/data/Pythia_LHEinput.cmd
```diff
@@ -7,10 +7,8 @@
 ! Adjusted from Pythia example: main42.cmnd
 
 ! 1) Settings that will be used in a main program.
-Main:numberOfEvents = 1            ! number of events to generate
-Main:timesAllowErrors = 3          ! abort run after this many flawed events
-Random:seed = 1234                ! initialize random generator with a seed
-
+Main:numberOfEvents = 100          ! number of events to generate
+Main:timesAllowErrors = 100        ! abort run after this many flawed events
```
clementhelsens: hum, 100 is maybe a bit too much no?
In a typical job of 10k events we don't expect more than a few failures.

jlingema:
ok to test.

Generation/data/Pythia_LHEinput.cmd
jlingema: Does this need to be changed or is it ok to merge?

selvaggi: no let's keep 100, does not harm. That's only 1% failure in 10k events

clementhelsens: ok, indeed it is more safe

PR #119, opened 2016-07-21
ACTS Integration (Tracking Geometry Service)

vvolkl:
- @jhrdinka's Tracking Geometry service based on the ACTS DD4hep plugin
- TrkGeoToGDMLDump Service for validation
- Algorithm that wraps current ACTS fitter example
- see https://gitlab.cern.ch/acts/a-common-tracking-sw/wikis/home for details
- requires an ACTS installation in `CMAKE_PREFIX_PATH` and a C++14 installation of DD4hep

Seeding and Hit conversion in future PR.

PR #171, opened 2017-02-22
Substitute PrintOverlapsSvc by Functioning Test

vvolkl:
The check on overlaps we had using PrintOverlapsSvc had a (yet to be understood) issue connected with using multiple calls to `loadXml` to build the detector. Since the point of the service was to allow using multiple detector compact files, and this is not allowed, I removed the service and substituted it by a test using the DD4hep python script which does the same thing. The tests are still failing because of overlooked overlaps -- for the tracker these are fixed with the acts PR #119.


Detector/DetFCChhBaseline1/compact/FCChh_DectMaster.xml
```diff
@@ -31,7 +31,7 @@
   <include ref="./FCChh_Shielding.xml" />
 
 
-  <include ref="../../DetFCChhTrackerSimple/compact/Tracker.xml" />
+  <include ref="../../DetFCChhTrackerTkLayout/compact/Tracker.xml" />
```
jlingema: Is the "Simple" tracker actually still needed? As far as I understand it has overlaps as it stands? If that's the case, I think we should either remove it or fix the overlaps.

vvolkl: I don't think it is still needed, for most purposes `Detector/DetFCChhBaseline1/compact/estimates/FCChh_TrackerMaterial.xml` is probably more useful. I'm working on a simplified version of the TkLayoutTracker that can be used in the reco also. I'll remove it.

vvolkl:
Overlaps in Tracker fixed in #119.

jlingema:
Sorry, I had set up Jenkins to run the tests in parallel, but that depends on #186 (and why the tests failed). I reverted the change on Jenkins until that one is merged.

retest this please

vvolkl:
retest this please

jlingema:
Thanks, Valentin.

PR #173, opened 2017-03-03
Restructure generation

vvolkl:
This PR removes the separate HepMCReader-, Pythia8- and ParticleGun Algorithms, in favor of a more centralised `GenAlg`, changing some of those components to tools in the process. The `GenAlg` has separate tools for signal- and pileup-hepmc input, making the pileup handling more flexible. It also adds the possibility to merge generation edm data after conversion.


Generation/src/components/PileUpEdmFileReaderTool.cpp
```diff
@@ -0,0 +1,81 @@
...
+  declareProperty("filename", m_filename = "input.root", "EDM file which contains the particle collections");
+  declareProperty("particleCollectionName", m_particleCollectionName = "allGenParticles");
+  declareProperty("vertexCollectionName", m_vertexCollectionName = "allGenVertices");
+}
+
+PileUpEdmFileReader::~PileUpEdmFileReader() { ; }
+
+StatusCode PileUpEdmFileReader::initialize() {
+  m_store = podio::EventStore();
+  m_reader = podio::ROOTReader();
```
jlingema: I don't like this usage of the reader and event store since it circumvents all the handles and related functionality of Gaudi. I think this needs some further discussion (also with @hegner).

Generation/src/components/MomentumRangeParticleGun.cpp
```diff
@@ -9,12 +9,7 @@
 #include "GaudiKernel/PhysicalConstants.h"
 #include "GaudiKernel/IRndmGenSvc.h" 
 
-//===========================================================================
-// Implementation file for class: MomentumRangeParticleGun
-//
-// 2008-05-18: Patrick Robbe adaptation to tool structure
-// 2014-06-22: Benedikt Hegner adaption for non LHCb use cases
-//===========================================================================
```
vvolkl: If we keep it, we should probably make it at least doxygen-readable. Do you maybe know of a suitable command to track a change + corresponding author?

Generation/src/components/HepMCFileWriter.h
```diff
@@ -0,0 +1,35 @@
...
+#include "GaudiAlg/GaudiAlgorithm.h"
+
+#include "FWCore/DataHandle.h"
+
+namespace HepMC {
+  class GenEvent;
+  class IO_GenEvent;
+}
+
+class HepMCFileWriter: public GaudiAlgorithm {
```
vvolkl: Sure, but then I think we should lead by example and stop storing store our samples in this format ;-)

Generation/src/components/PileUpEdmFileReaderTool.cpp
vvolkl: Ah yes, after our discussion I meant to remove it. I implemented a very un-generic algorithm instead:

https://github.com/HEP-FCC/FCCSW/pull/173/commits/a791b2edfeefc4a56010f93c53a4daa7bb2a7f4a#diff-f363a567fa59d2cf5e319fc5297de46a

But this way has the drawback of being way more flexible (all events need to be merged before simulation). Yes, we should discuss.

Generation/src/components/HepMCFileWriter.h
jlingema: It's only test samples that we have in HepMC which is sort of in line with my statement about "debugging" purposes ;-). We can certainly discuss whether those tests using these files should maybe be moved outside of the Examples.

Generation/src/components/MomentumRangeParticleGun.cpp
jlingema: If you want to modify it to be doxygen-compliant, fine. You can just list the authors and put the changes behind them? E.g.:
```c++
/**
  @author Patrick Robbe (adaptation to tool structure)
  @author Benedikt Hegner (adaption for non LHCb use cases)
  @date 2008-2016
*/
```

vvolkl: Oh, I remember now why I removed it. It was already as a doxygen comment in the header (where it belongs, I guess) I updated it with the changes.

Generation/src/components/GenAlg.cpp
```diff
@@ -22,40 +19,34 @@ HepMCReader::HepMCReader(const std::string& name, ISvcLocator* svcLoc):
   declareProperty("HepMCMergeTool", m_HepMCMergeTool);
   declarePublicTool(m_HepMCMergeTool, "HepMCSimpleMerge/HepMCMergeTool");
 
-  declareProperty("FileReaderSignal", m_signalFileReader);
-  declarePrivateTool(m_signalFileReader, "HepMCFileReader/FileReaderSignal");
-  declareProperty("FileReaderPileUp", m_pileupFileReader);
-  declarePrivateTool(m_pileupFileReader, "HepMCFileReader/FileReaderPileup");
+  declareProperty("SignalProvider", m_signalProvider);
+  declarePublicTool(m_signalProvider, "MomentumRangeparticleGun/HepMCProviderTool");
```
vvolkl: Done

Generation/Generation/IParticleGunTool.h
```diff
@@ -13,7 +14,7 @@
  *  @date   2008-05-18
  */
 
-class IParticleGunTool : virtual public IAlgTool {
+class IParticleGunTool : virtual public IAlgTool, virtual public IHepMCProviderTool {
```
vvolkl: Done

Generation/src/components/MomentumRangeParticleGun.cpp
jlingema: Ah, my bad.. Missed that it is in the cpp.

Generation/src/components/GenMerge.cpp
```diff
@@ -0,0 +1,64 @@
...
+  auto collPSig =  m_part_InSignal.get();
+
+  auto collVPil = m_vert_InPileUp.get();
+  auto collPPil =  m_part_InPileUp.get();
+
+  fcc::MCParticleCollection* collPOut = new fcc::MCParticleCollection();
+  fcc::GenVertexCollection* collVOut = new fcc::GenVertexCollection();
+
+  for(const auto elem: *collVSig) {
+    collVOut->push_back(elem.clone());
```
jlingema: As we just discussed, you'll also need to update the relations of the vertices here. Otherwise, if the signal / pile-up collection is dropped this will lead to inconsistent relations.

Generation/src/components/MomentumRangeParticleGun.cpp
```diff
@@ -106,35 +106,29 @@ void MomentumRangeParticleGun::generateParticle(Gaudi::LorentzVector& momentum,
...
+      1);  // hepmc status code for final state particle
 
-  v -> add_particle_out( p ) ;
+  v->add_particle_out(p);
 
-  theEvent.add_vertex( v ) ;
-  theEvent.set_signal_process_id( 0 ) ;
-  theEvent.set_signal_process_vertex( v ) ;
+  theEvent.add_vertex(v);
+  theEvent.set_signal_process_id(0);
```
jlingema: Should this be hard-coded to 0?

Generation/src/components/MomentumRangeParticleGun.h
```diff
@@ -12,8 +12,8 @@
  *
  *  Particle gun with given momentum range
  *
- *  @author Patrick Robbe
- *  @author Benedikt Hegner
+ *  @author Patrick Robbe (adaptation to tool structure)
```
jlingema: The original author adapted to tool structure? I guess that this was a mistake and you wanted to add that to your own name, which is missing from the list?

Generation/src/components/EDMConverter.cpp
```diff
@@ -0,0 +1,58 @@
+#include "EDMConverter.h"
```
vvolkl: Ok.  Also `HepMCConverter` -> `HepMCToEDMConverter`?

jlingema: Hm, probably best to stick to a theme.. So yes.

Generation/src/components/MomentumRangeParticleGun.cpp
vvolkl: Good point. I think 0 is a good value as any for a particle gun, but for a default-constructed HepMC::GenEvent it is 0 anyway, so I think we can remove this line, just for the rare case in which the gun is used together with something else which sets it.

Generation/src/components/MomentumRangeParticleGun.h
vvolkl: No, this was the comment that was originally in the src file. I guess "tool structure" can mean Gaudi.

Generation/src/components/GenMerge.h
```diff
@@ -0,0 +1,45 @@
...
+
+private:
+  DataHandle<fcc::GenVertexCollection> m_vert_InSignal{"signalGenParticles", Gaudi::DataHandle::Reader, this};
+  DataHandle<fcc::MCParticleCollection> m_part_InSignal{"signalGenVertices", Gaudi::DataHandle::Reader, this};
+
+  DataHandle<fcc::GenVertexCollection> m_vert_InPileUp{"pileupGenVertices", Gaudi::DataHandle::Reader, this};
+  DataHandle<fcc::MCParticleCollection> m_part_InPileUp{"pileupGenParticles", Gaudi::DataHandle::Reader, this};
+
+  DataHandle<fcc::GenVertexCollection> m_vert_Out{"allGenVertices", Gaudi::DataHandle::Writer, this};
+  DataHandle<fcc::MCParticleCollection> m_part_Out{"allGenParticles", Gaudi::DataHandle::Writer, this};
```
vvolkl: Done

Generation/src/components/HepMCFileWriter.cpp
```diff
@@ -0,0 +1,32 @@
+#include "HepMCFileWriter.h"
+#include "HepMC/GenEvent.h"
+#include "HepMC/IO_GenEvent.h"
+
+DECLARE_COMPONENT(HepMCFileWriter)
+
+HepMCFileWriter::HepMCFileWriter(const std::string& name, ISvcLocator* svcLoc) : GaudiAlgorithm(name, svcLoc) {
+  declareProperty("Filename", m_filename = "Output_HepMC.dat", "Name of the HepMC file to write");
```
vvolkl: Done

Generation/src/components/EDMConverter.cpp
```diff
@@ -0,0 +1,58 @@
...
+  // First cover the case that hepMC file is not in expected units and then convert to EDM default
+  double hepmc2EdmLength =
+      HepMC::Units::conversion_factor(event->length_unit(), gen::hepmcdefault::length) * gen::hepmc2edm::length;
+  double hepmc2EdmEnergy =
+      HepMC::Units::conversion_factor(event->momentum_unit(), gen::hepmcdefault::energy) * gen::hepmc2edm::energy;
+
+  for (auto p : *(particles)) {
+    if (p.status() == 1) {  // only final state particles
+      HepMC::GenParticle* pHepMC =
+          new HepMC::GenParticle(HepMC::FourVector(p.p4().px, p.p4().py, p.p4().pz, p.p4().mass / hepmc2EdmEnergy),
```
vvolkl: Done

PR #174, opened 2017-03-07
refined HCAL geometry

CoralieNeubueser:


jlingema:
As discussed in private email, this needs a rebase to remove files we don't want in the master.

PR #177, opened 2017-03-09
Remove remaining afs dependencies

jlingema:
I moved all test samples into our EOS project space for FCCSW. I did not move them on the actual EOS data volume to avoid having to setup the client and management element whenever you want to test. (The project space is mounted by default on lxplus and the Jenkins build nodes.)


zaborowska:
Hi Joschka,

how about our Pythia examples - they contain links to Les Houches Events on afs, so maybe it is worth moving them to eos as well? [here](https://github.com/jlingema/FCCSW/blob/eos/Generation/data/Pythia_pp_zgzg_4l.cmd#L36) and [here](https://github.com/jlingema/FCCSW/blob/eos/Generation/data/Pythia_pp_h_4l.cmd#L36)

And to remove **all** afs dependencies, could you also change SAS and llvm paths in
 [./checkformat](https://github.com/jlingema/FCCSW/blob/eos/checkformat#L8)?

jlingema:
Hi Anna,

The checkformat is already fixed in #175. As for the lhe files, I wanted to chat with @selvaggi before moving those into EOS. 

Cheers,
Joschka

jlingema:
Had a chat with him and we agreed to put them in the same space as the other test samples.

zaborowska:
Ah, sorry, I didn't notice it there.
Thanks a lot!

PR #178, opened 2017-03-15
Detector dimensions

faltovaj:
Hi everyone,
I changed the FCChh detector dimensions in ECAL + HCAL to follow the recent geometry from Werner. The barrel stays almost the same (just HCAL rmax changed a bit, it is agreed with Werner), but there are changes in the extended barrel and forward. ( @jlingema , may I ask you to double check the numbers? Thanks! ) There are parts with a common envelope for ECAL + HCAL (e.g. forward region has a common cryostat for these two). The question is what to do with the system id in this case. I reserved a place for 2 IDs in these cases. Let me know if you can come up with something smarter.

I changed EcalAir.xml + HcalAir.xml to follow the new naming convention.

The question is what to do with the Ecal + Hcal material estimates. These would require some more changes to be useful. The question is if it is really needed. We hope to add the full geometry description in the extended barrels & forward at some point. @clementhelsens , what do you think? Can we just remove these?

### DIFF
--- a/Detector/DetFCChhBaseline1/compact/FCChh_DectDimensions.xml
+++ b/Detector/DetFCChhBaseline1/compact/FCChh_DectDimensions.xml
@@ -65,50 +65,47 @@
     <constant name="FwdTracker_z2" value="2000*mm"/>
     <constant name="FwdTracker_zOffset" value="14000*mm"/>
 
-    <!-- E(M)CAL DIMENSIONS -->
-    <!-- E(M)Cal Barrel -->
+    <!-- CALORIMETER DIMENSIONS -->
+    <!-- ECal Barrel (EM B) -->
     <constant name="BarECal_id" value="5"/>
     <constant name="BarECal_rmin" value="1750*mm"/>
     <constant name="BarECal_rmax" value="2750*mm"/>
     <constant name="BarECal_dz" value="5000*mm"/>
-    <!-- E(M)Cal End-caps -->
-    <constant name="EndCapECal_id" value="6"/>
-    <constant name="EndCapECal_rmin1" value="864.278*mm"/>
-    <constant name="EndCapECal_rmax1" value="2800*mm"/>
-    <constant name="EndCapECal_rmin2" value="1467.641*mm"/>
-    <constant name="EndCapECal_rmax2" value="2800*mm"/>
-    <constant name="EndCapECal_dz" value="1850*mm"/>
-    <constant name="EndCapECal_zOffset" value="5300*mm + EndCapECal_dz"/>
-    <!-- E(M)Cal forward -->
-    <constant name="FwdECal_id" value="7"/>
-    <constant name="FwdECal_rmin" value="30*mm"/>
-    <constant name="FwdECal_rmax" value="3500*mm"/>
-    <constant name="FwdECal_dz" value="750*mm"/>
-    <constant name="FwdECal_zOffset" value="16500*mm+FwdECal_dz"/>
-
-
-    <!-- HCAL DIMENSIONS -->
-    <!-- HCal Barrel -->
+    <!-- HCal Barrel (HCAL B) -->
     <constant name="BarHCal_id" value="8"/>
     <constant name="BarHCal_rmin" value="2850*mm"/>
     <constant name="BarHCal_rmax" value="4850*mm"/>
     <constant name="BarHCal_dz" value="4600*mm"/>
-    <!-- HCal End-Cap -->
-    <constant name="EndCapHCal_id" value="9"/>
-    <constant name="EndCapHCal_rmin" value="2850*mm"/>
-    <constant name="EndCapHCal_rmax" value="4550*mm"/>
-    <constant name="EndCapHCal_dz" value="2000*mm"/>
-    <constant name="EndCapHCal_zOffset" value="5000*mm + EndCapHCal_dz"/>
-    <!-- HCal forward -->
-    <constant name="FwdHCal_id" value="10"/>
-    <constant name="FwdHCal_rmin" value="30*mm"/>
-    <constant name="FwdHCal_rmax" value="3500*mm"/>
-    <constant name="FwdHCal_dz" value="1250*mm"/>
-    <constant name="FwdHCal_zOffset" value="18000*mm + FwdHCal_dz"/>
+    <!-- Extended barrel (EM EB id 6 + HEC id 7)-->
+    <constant name="ExtBarCal_id" value="6"/>
+    <constant name="ExtBarCal_rmin1" value="776*mm"/>
+    <constant name="ExtBarCal_rmax1" value="2700*mm"/>
+    <constant name="ExtBarCal_rmin2" value="1272*mm"/>
+    <constant name="ExtBarCal_rmax2" value="2700*mm"/>
+    <constant name="ExtBarCal_dz" value="1500*mm"/>
+    <constant name="ExtBarCal_zOffset" value="5300*mm + ExtBarCal_dz"/>
+    <!-- HCal extended barrel (HCAL EB) -->
+    <constant name="ExtBarHCal_id" value="9"/>
+    <constant name="ExtBarHCal_rmin1" value="3500*mm"/>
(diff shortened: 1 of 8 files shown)


zaborowska:
ok to test

clementhelsens:
Hello Jana, 

By "these" you mean the extended and forward envelops? 
For the material scan, this is a very important tool to get a first validation of the detector implementation. Why does it requires more changes to be useful?

Cheers,
Clement

faltovaj:
Hi Clement,
I mean the material estimates of the extended barrels and the forward calorimeters, e.g. https://github.com/HEP-FCC/FCCSW/blob/master/Detector/DetFCChhBaseline1/compact/estimates/FCChh_ECalFwdMaterial.xml
In this example the whole envelope is filled with the mixture of active+passive material, which is not true. There is a cryostat material, LAr bath, different materials in ECAL & HCAL missing. My question is if this estimate is really needed when we plan to implement the real geometry description soon.
Cheers,
Jana

Detector/DetFCChhBaseline1/compact/FCChh_DectDimensions.xml
```diff
@@ -65,58 +65,55 @@
...
-    <constant name="FwdECal_zOffset" value="16500*mm+FwdECal_dz"/>
-
-
-    <!-- HCAL DIMENSIONS -->
-    <!-- HCal Barrel -->
+    <!-- HCal Barrel (HCAL B) -->
     <constant name="BarHCal_id" value="8"/>
     <constant name="BarHCal_rmin" value="2850*mm"/>
-    <constant name="BarHCal_rmax" value="4850*mm"/>
+    <constant name="BarHCal_rmax" value="4890*mm"/>
```
faltovaj: Hi Joschka, that's exactly it. It is a new requirement from HCAL, as the support is a bit bigger than we thought it would be.

Detector/DetFCChhBaseline1/compact/FCChh_DectDimensions.xml
```diff
@@ -65,58 +65,55 @@
...
+    <constant name="ExtBarCal_rmin1" value="776*mm"/>
+    <constant name="ExtBarCal_rmax1" value="2700*mm"/>
+    <constant name="ExtBarCal_rmin2" value="1272*mm"/>
+    <constant name="ExtBarCal_rmax2" value="2700*mm"/>
+    <constant name="ExtBarCal_dz" value="1500*mm"/>
+    <constant name="ExtBarCal_zOffset" value="5300*mm + ExtBarCal_dz"/>
+    <!-- HCal extended barrel (HCAL EB) -->
+    <constant name="ExtBarHCal_id" value="9"/>
+    <constant name="ExtBarHCal_rmin1" value="3500*mm"/>
+    <constant name="ExtBarHCal_rmax1" value="4590*mm"/>
```
faltovaj: That's the same change as before for the HCAL barrel - 4 cm more.

Detector/DetFCChhBaseline1/compact/FCChh_DectDimensions.xml
```diff
@@ -65,58 +65,55 @@
...
+    <constant name="ExtBarHCal_rmin2" value="2850*mm"/>
+    <constant name="ExtBarHCal_rmax2" value="4590*mm"/>
+    <constant name="ExtBarHCal_dz1" value="150*mm"/>
+    <constant name="ExtBarHCal_zOffset1" value="5000*mm + ExtBarHCal_dz1"/>
+    <constant name="ExtBarHCal_dz2" value="1350*mm"/>
+    <constant name="ExtBarHCal_zOffset2" value="5300*mm + ExtBarHCal_dz2"/>
+    <!-- Forward calorimeters (EFCAL id 10 + HFCAL id 11) -->
+    <constant name="FwdCal_id" value="10"/>
```
faltovaj: Sure, I will change this. How easy solution :)

Detector/DetFCChhBaseline1/compact/FCChh_DectDimensions.xml
```diff
@@ -65,58 +65,55 @@
     <constant name="FwdTracker_z2" value="2000*mm"/>
     <constant name="FwdTracker_zOffset" value="14000*mm"/>
```
jlingema: I noticed that also the forward part of the tracker changed in the meantime, I guess this should be:

```xml
     <!-- single layer close to ExCal -->
     <constant name="FwdTracker_rmin1" value="40*mm"/>
     <constant name="FwdTracker_rmax1" value="1173*mm"/>
     <constant name="FwdTracker_z1" value="7500*mm"/>
     <!-- the rest of the tracking layers inside the forward solenoid / shielding -->
     <constant name="FwdTracker_rmin2" value="60*mm"/>
     <constant name="FwdTracker_rmax2" value="1556*mm"/>
      <constant name="FwdTracker_z2" value="3000*mm"/>
      <constant name="FwdTracker_zOffset" value="13000*mm"/>
```

Can you please double-check @vvolkl / @drasal?

vvolkl: Hi! @drasal, seems like we are currently using a `zOffset` of 10000 mm and a `z2` of 6000 mm. I don't know which one is the most recent config!

jlingema: The offset points to the center and the length is a half-length so that looks OK. What about the single layer closer to the IP?

drasal: Hi,

the FwdHelp (intermediate forward tracker between the central tracker and forward tracker) is described here: http://fcc-tklayout.web.cern.ch/fcc-tklayout/FCChh_Option3.v02/indexFwdHelp.html
The parameters are: rMin = 40mm rMax = 1189.4, zMean = 7500

The Fwd tracker can be in detail found here: http://fcc-tklayout.web.cern.ch/fcc-tklayout/FCChh_Option3.v02/indexFwd.html

jlingema: Hi! Thanks for the numbers. A couple of small detail questions:
- Why has rMax changed (out of curiosity)?
- Technically the forward tracker extrudes from the envelope given by Werner by the width of one forward disk
Can you make sure the corresponding correct numbers end up in Werner's "master document"?

drasal: Hi Joschka, 

* There are several reasons, why the rMax changed in the forward region. I've realized these after the FCC-hh meeting, when several clever comments were made... First, I changed the modules sizes to be the same as in barrel region (the less different types of modules, the lower price). In addition, with such size and given pitch, we have 256 or 512 or 1024 read-out channels (dividable by 2), which is what one in reality wants. We can then easily estimate the number of needed chips, which affects the total cost. After recalculation of all positions to have a hermetic detector, one comes up with a different rMax. Second, Werner always wanted the detectors to end-up at r=155cm, with a new module size I'm closer to that value...
* If there is a space of 10-15cm for services above the helpFwd disc (so between the fwd disc and the calorimeter), this shouldn't be a problem.
* Sure, I will propagate them. Good point!

jlingema: I finally implemented the numbers and indeed see a clash between the extended barrel calorimeters and the "FwdHelp": The calorimeters themselves go down to eta 2.5 but then there is 10 cm of cryo foreseen below that. The lower radius at z = 7.5 m of the cryo system corresponds to 1.223 m.

drasal: I have fixed this issue by removing one ring in "FwdHelp". The new geometry is available at: http://fcc-tklayout.web.cern.ch/fcc-tklayout/FCChh_Option3.v02/index.html

jlingema: OK, thanks. Updated #182 accordingly.

PR #182, opened 2017-03-21
Updating FCC-hh detector dimensions

jlingema:
This supersedes PR #178 adapting also the material estimates and updating numbers for shielding and the forward tracker.


Detector/DetFCChhBaseline1/compact/FCChh_DectDimensions.xml
```diff
@@ -57,66 +57,66 @@
...
-    <constant name="FwdTracker_rmin2" value="2000*mm"/>
-    <constant name="FwdTracker_rmax2" value="3505*mm"/>
-    <constant name="FwdTracker_z2" value="2000*mm"/>
-    <constant name="FwdTracker_zOffset" value="14000*mm"/>
-
-    <!-- E(M)CAL DIMENSIONS -->
-    <!-- E(M)Cal Barrel -->
+    <!-- single layer close to ExCal -->
+    <constant name="FwdTracker_rmin1" value="40*mm"/>
+    <constant name="FwdTracker_rmax1" value="1173*mm"/>
```
jlingema: This number still needs to be updated according to the discussion in #178

Detector/DetFCChhBaseline1/compact/FCChh_DectDimensions.xml
```diff
@@ -57,66 +57,66 @@
...
+    <constant name="ExtBarCal_rmin2" value="1272*mm"/>
+    <constant name="ExtBarCal_rmax2" value="2700*mm"/>
+    <constant name="ExtBarCal_dz" value="1500*mm"/>
+    <constant name="ExtBarCal_zOffset" value="5300*mm + ExtBarCal_dz"/>
+    <!-- HCal extended barrel (HCAL EB) -->
+    <constant name="ExtBarHCal_id" value="9"/>
+    <constant name="ExtBarHCal_rmin1" value="3500*mm"/>
+    <constant name="ExtBarHCal_rmax1" value="4590*mm"/>
+    <constant name="ExtBarHCal_rmin2" value="2850*mm"/>
+    <constant name="ExtBarHCal_rmax2" value="4590*mm"/>
```
zaborowska: I actually don't like it being called the same way (_rmin1 _rmax2 etc) as it is for Extended barrel above. The shapes are very different (trapezoid vs. two rectangles).
Wouldn't it be better called ExtBarHCal1_rmin or ExtBarHCalpart1_rmin etc. ? Same for dz below.

Detector/DetFCChhBaseline1/compact/FCChh_DectDimensions.xml
```diff
@@ -57,66 +57,66 @@
...
+    <constant name="ExtBarHCal_dz1" value="150*mm"/>
+    <constant name="ExtBarHCal_zOffset1" value="5000*mm + ExtBarHCal_dz1"/>
+    <constant name="ExtBarHCal_dz2" value="1350*mm"/>
+    <constant name="ExtBarHCal_zOffset2" value="5300*mm + ExtBarHCal_dz2"/>
+    <!-- Forward calorimeters (EFCAL + HFCAL) -->
+    <constant name="FwdECal_id" value="10"/>
+    <constant name="FwdHCal_id" value="11"/>
+    <constant name="FwdCal_rmin1" value="49.6*mm"/>
```
zaborowska: 16.5 * tan(Theta(eta=6)) = 81.8 mm

Detector/DetFCChhBaseline1/compact/FCChh_DectDimensions.xml
```diff
@@ -57,66 +57,66 @@
...
+    <constant name="FwdCal_rmin2" value="58.6*mm"/>
```
zaborowska: 19.5*tan(Theta(eta=6)) = 96.7 mm

Detector/DetFCChhBaseline1/compact/FCChh_DectDimensions.xml
```diff
@@ -57,66 +57,66 @@
...
-
-    <!-- E(M)CAL DIMENSIONS -->
-    <!-- E(M)Cal Barrel -->
+    <!-- single layer close to ExCal -->
+    <constant name="FwdTracker_rmin1" value="40*mm"/>
+    <constant name="FwdTracker_rmax1" value="1076.8*mm"/>
+    <constant name="FwdTracker_zOffset1" value="7500*mm"/>
+    <!-- the rest of the tracking layers inside the forward solenoid / shielding -->
+    <constant name="FwdTracker_rmin2" value="60*mm"/>
+    <constant name="FwdTracker_rmax2" value="1556*mm"/>
```
drasal: The R-max value is 1549.0 ...

Detector/DetFCChhBaseline1/compact/FCChh_DectDimensions.xml
```diff
@@ -57,66 +57,66 @@
...
+    <constant name="FwdTracker_z2" value="3000*mm"/>
```
drasal: If these values are a problem, I may fit the fwd tracker into volume between 10-16m. The average disc values are at 10 and 16m, respectively.

Detector/DetFCChhBaseline1/compact/FCChh_DectDimensions.xml
jlingema: OK, will change it accordingly

Detector/DetFCChhBaseline1/compact/FCChh_DectDimensions.xml
jlingema: Nope, not a problem, we'd just like to keep things consistent

jlingema:
Thanks Anna. The commits I just pushed should address all your comments. Let me know if I missed something :-)

Detector/DetFCChhBaseline1/compact/FCChh_DectDimensions.xml
zaborowska: Hi Zbynek, so what is the disk Radius-max?

zaborowska:
Thanks Zbynek for checking.
So the main idea is that those numbers here should describe the envelopes, so every element of the detector should be contained inside, that's why I requested that change in fwd tracker.

drasal:
So, the rMax of the FWD part is 1549.8mm. There is one comment to the Z envelope of the central tracker, currently it's up-to 5000mm, but tkLayout has it up-to 5009.7mm. Similarly, the Z envelope of the fwd tracker should extend from 9990.4mm to 16009.7mm 

If there is any clash with other detectors, let me know and I will force tkLayout to have the envelope on these values, not the middle positions at these values.

Detector/DetFCChhBaseline1/compact/FCChh_DectDimensions.xml
drasal: 1549.8mm

drasal:
Just for completeness, the FwdHelp disc extends from zMin=7490.4 to zMax=	7509.6, the rMax=1076.8mm

jlingema:
So the barrel numbers should also be updated, I guess? Are your numbers being put in the next version of Werner's document, Zbynek?

drasal:
Yes, I have to send all these numbers to Werner's document. Actually, these numbers didn't change, Werner might have taken the avg disc Z position, which is 5m for example for the central tracker. Hence, the discrepancy.

drasal:
Is the tracker the only detector, which differs from Werner's document?

jlingema:
Hi Zbynek,

No, it wasn't only the tracker. Also the HCal design differs from what is in the document. The rest of the changes were due to an old version used when I first implemented the numbers. 

Since the envelopes be actually enveloping, as Anna already pointed out, I'll also update the central tracker.

Thanks,
Joschka

drasal:
Joschka, thanks for your update of the central tracker. Just to be sure, it doesn't collide with the ECAL, correct?

jlingema:
No collisions, no. The extended barrel of the ECal starts at 5.3 m, so the small change (5 m -> 5.0097 m) also doesn't really change anything w.r.t. space for services, support, etc.

Issue #184, opened 2017-03-28
Cleaning tests


jlingema:
Some of the configurations used for testing are redundant since there are examples that use all included components.

Other parts of the software are not covered, e.g. not all "generic shapes" in DetCommon are tested, corresponding configs should be added.

PR #186, opened 2017-03-30
Bringing tests up to date, minor clean up

jlingema:
Addresses #184.

- Adding tests for all primitive shapes in `Detector/DetCommon`
- Removing redundant tests (not too many)
- Cleaning up FWCore (moving source files according to conventions)
- Fix some dependencies in tests (now allows to run the tests in parallel with `make test ARGS="-j 8"`
- Removing remnant files from old tracking code


vvolkl:
Ah, there also seems to be no test that uses the master detector xml, but that's addressed in the overlap-check PR, I think.

jlingema:
Thanks for the comments. I removed the test as you suggest. We already have tests that either [use the magnetic field](https://github.com/HEP-FCC/FCCSW/blob/master/Examples/options/geant_fullsim_field.py) or [don't](https://github.com/HEP-FCC/FCCSW/blob/master/Examples/options/geant_fullsim.py). Am I missing something?
As you already say, the master will be used in the overlap check, so I think that's fine.

PR #187, opened 2017-04-05
FCCee first import of drivers and xml files

Voutsi:
Addresses # .
FCCee_o3_v03 and FCCee_o4_v01 models.
Uses the lcgeo drivers with minor modifications + few new drivers for the new model

-
-

To be done:

- [ add ACTS extensions for tracking]
- [ ]

### DIFF
--- a/Detector/DetFCCeeBaseline1/CMakeLists.txt
+++ b/Detector/DetFCCeeBaseline1/CMakeLists.txt
@@ -1,0 +1,26 @@
+################################################################################
+# Package: DetFCCeeBaseline1
+################################################################################
+gaudi_subdir(DetFCCeeBaseline1 v1r0)
+
+gaudi_depends_on_subdirs(GaudiKernel Detector/DetCommon)
+
+
+find_package(DD4hep REQUIRED COMPONENTS DDCore DDSegmentation DDRec DDG4)
+find_package(Geant4)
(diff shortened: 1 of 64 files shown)


jlingema:
Hi,

Many thanks for the pull-request. I have a couple of general comments before starting to have a look at the code.

- There seem to be two versions of the detector layout, `FCCee_03_v03` and `FCCee_04_v01`, I assume that we either only want one or need to move them into separate baseline directories, see also [this paragraph](https://github.com/HEP-FCC/FCCSW/blob/master/Detector/doc/DD4hepInFCCSW.md#fccsw-folder-structure) in the detector doc
- At a cursory glance I spotted that there are several `material.xml`, `elements.xml` and `Material_v01.xml` can you please move to the ones that we have already in FCCSW [here](https://github.com/HEP-FCC/FCCSW/tree/master/Detector/DetCommon/compact)
- The idea was to have separate folders / sub-packages for the individual sub-detectors (also to allow exchanging different designs in a straight forward manner), I see that you have both drivers and xml descriptions all in the main directories, could you please spread them according to the doc mentioned above?
- The files are currently versioned with post-fixes `vXY`. Our approach was to instead use the code versioning itself. Can you drop the post-fixes or let us know why this approach is needed?

Thanks,
Joschka

jlingema:
Superseded by #196

PR #189, opened 2017-04-05
Average material over phi in Material Scan, fix PE material description

vvolkl:
Adds support for a materialscan averaged over randomly drawn values of phi for the eta-slices. Also fixes a bug in the material description, where the `fraction` and `compound ` descriptions were mixed up.

### DIFF
--- a/Detector/DetCommon/compact/materials.xml
+++ b/Detector/DetCommon/compact/materials.xml
@@ -141,8 +141,8 @@
 
   <material name="PE">
     <D type="density" value="0.95" unit="g/cm3"/>
-    <fraction n="4" ref="H"/>
-    <fraction n="2" ref="C"/>
+    <fraction n="0.6666666" ref="H"/>
+    <fraction n="0.3333333" ref="C"/>
   </material>
 
   <material name="Kapton">
--- a/Detector/DetComponents/src/MaterialScan.cpp
+++ b/Detector/DetComponents/src/MaterialScan.cpp
@@ -3,6 +3,8 @@
 
 #include "GaudiKernel/ITHistSvc.h"
 #include "GaudiKernel/Service.h"
+#include "GaudiKernel/RndmGenerators.h"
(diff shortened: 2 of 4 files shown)


Detector/DetCommon/compact/materials.xml
```diff
@@ -141,8 +141,8 @@
 
   <material name="PE">
     <D type="density" value="0.95" unit="g/cm3"/>
-    <fraction n="4" ref="H"/>
-    <fraction n="2" ref="C"/>
+    <fraction n="0.6666666" ref="H"/>
+    <fraction n="0.3333333" ref="C"/>
```
jlingema: since we create release notes from the headline of the PR, can you add this, since it may affect others?

Detector/DetComponents/src/MaterialScan.cpp
```diff
@@ -17,6 +17,7 @@ MaterialScan::MaterialScan(const std::string& name, ISvcLocator* svcLoc) : Servi
   declareProperty("filename", m_filename, "file name to save the tree to");
   declareProperty("etaBinning", m_etaBinning, "eta bin size");
   declareProperty("etaMax", m_etaMax, "maximum eta value");
+  declareProperty("nPhiTrials", m_nPhiTrials, "number of random phi values to average over");
```
jlingema: I seem to have missed these properties, when I moved everything to the new Gaudi. Can you do it (I can also do it after we merge)?

Detector/DetComponents/src/MaterialScan.h
```diff
@@ -28,4 +29,8 @@ class MaterialScan : public Service {
   double m_etaBinning;
   /// Maximum eta until which to scan, scan is performed from -m_etaMax to +m_etaMax
   double m_etaMax;
+  /// number of different phi values to average over
```
jlingema: can you add to the comment how the phi values are chosen?

PR #190, opened 2017-04-08
updated Delphes cards, added jet substructure

selvaggi:
Addresses # .

- updated baseline delphes FCChh card, including latest tracker and calorimeter parameterisations, and fatjet/jet substructure observables
- included interface to jet substructure

To be done:

- waiting for update on ECAL resolution to be included


selvaggi:
tests are probably failing because this commit requires the following change in fcc-edm:
https://github.com/HEP-FCC/fcc-edm/pull/46

jlingema:
Hi Michele,

With the Full-Sim detector description we went the route of just keeping one version (as long as there is not two concepts that should be used in parallel). Is there a reason you added `v02` instead of changing `v01`?

Cheers,
Joschka

drasal:
Hello Michele and Joschka,

just a recommendation... If necessary, we may keep the history of individual Delphes cards on fcc-hh project workspace, which I've created on EOS...

jlingema:
I don't see why we need to keep the history also in files. Git keeps all versions that were committed. Why do we need to keep it twice? :-)

selvaggi:
Ok for me, we can leave only one version.

selvaggi:
Hi Joschka,  @clementhelsens,

we want to keep them also in FCCSW, it is easier for tests and tutorials. I will update the cards on EOS and in FCCSW at the same time. 

Cheers
Michele

clementhelsens:
Yes agreed! We need to decouple the tutorials/tests from production

jlingema:
Then we should also update the documentation to make this very clear. Otherwise, I fear, this can quickly lead to confusion and de-synchronised cards.

I would actually vote for having a separate, simple mock-up card just for the tests and point people to the production cards in the tutorials. But that's just my two cents.

clementhelsens:
yes, that's a good point, we already have the root files for the tutorials in eos (/eos/fcc/hh/tutorials/)
so we can also have the cards there. But we'll need to read cards from eos without copying them, and I don't know how to do that with text files yet

jlingema:
Can you rebase / merge with the master once PR #200 is merged? That will set up the updated edm with the needed data types.
Thanks!

Issue #163, opened 2017-01-31
[Sliding window] Sharing cells between clusters


zaborowska:
If clusters have cells in common, the energy of those cells is double counted.
Temporary solution: use duplicate removal window of the same size as the final window.
To do: find cells shared between clusters and add only a part of their energy to clusters.

zaborowska:
Fixed in #193

Issue #162, opened 2017-01-31
[Sliding window] Building towers


zaborowska:
Currently the tower for the sliding window must be a multiple of the cell size (in eta and phi).
To do: implement energy splitting of cell energy into reconstruction towers.

zaborowska:
Fixed in #193

PR #193, opened 2017-04-21
Segmentation offset

### LINKED ISSUE #161: [Sliding window] Deposits at detector borders
If the maximum of the cluster energy is close to the detector border, it may be omitted while calculating the barycentre position.
Result: In this case the position is shifted towards the centre of the detector. 
To fix: calculation of the barycentre position.

faltovaj:
Main changes & comments:
1/ Fixing the Sliding window caveats mentioned in the Issues (fixes #161, #162, #163)
2/ Adding an offset in the ECAL phi-eta segmentation. There are only non-negative phi/eta identifiers now. Makes life easier, especially when we will combine different calorimeters with different segmentations.
3/ A couple of improvements and fixes.

Let me know if you have any questions or comments :)

Thanks,
Jana


zaborowska:
ok to test

Reconstruction/RecCalorimeter/src/components/CreateCaloClustersSlidingWindow.cpp
```diff
@@ -213,18 +225,55 @@ StatusCode CreateCaloClustersSlidingWindow::execute() {
...
+      // loop over all clusters and check if they have any tower in common with our current cluster
+      for (const auto iclu : m_preClusters) {
+        int idEtaIcl = m_towerTool->idEta(iclu.eta);
+        int idPhiIcl = m_towerTool->idPhi(iclu.phi);
+        if (idEtaCl != idEtaIcl && idPhiCl != idPhiIcl) {
+          if (abs(idEtaIcl - idEtaCl) <= halfEtaFin && abs(idPhiIcl - idPhiCl) <= halfPhiFin) {
+            int iEtaShareMin = (idEtaIcl < idEtaCl) ? idEtaCl - halfEtaFin : idEtaIcl - halfEtaFin;
+            int iEtaShareMax = (idEtaIcl < idEtaCl) ? idEtaIcl + halfEtaFin : idEtaCl + halfEtaFin;
+            int iPhiShareMin = (idPhiIcl < idPhiCl) ? idPhiCl - halfPhiFin : idPhiIcl - halfPhiFin;
+            int iPhiShareMax = (idPhiIcl < idPhiCl) ? idPhiIcl + halfPhiFin : idPhiCl + halfPhiFin;
```
jlingema: I think this would be more readable with `std::min` / `std::max`. Also I don't get the meaning of the suffix `Icl` :)

Reconstruction/RecCalorimeter/src/components/SingleCaloTowerTool.cpp
```diff
@@ -41,96 +43,134 @@ StatusCode SingleCaloTowerTool::initialize() {
...
+  // Get the input collection with cells from simulation + digitisation (after
+  // calibration and with noise)
   const fcc::CaloHitCollection* cells = m_cells.get();
   debug() << "Input Hit collection size: " << cells->size() << endmsg;
   // Loop over a collection of calorimeter cells and build calo towers
-  int iPhi = 0, iEta = 0;
-  float etaCell = 0;
+  // borders of the cell in eta/phi
+  float etaCellMin, etaCellMax;
+  float phiCellMin, phiCellMax;
```
jlingema: Can you also initialise these variables with sensible defaults?

Reconstruction/RecCalorimeter/src/components/SingleCaloTowerTool.cpp
```diff
@@ -41,96 +43,134 @@ StatusCode SingleCaloTowerTool::initialize() {
...
+    fracPhiMin = 1.0;
+    fracPhiMax = 1.0;
+    fracPhiRest = 1.0;
+    if (iPhiMin != iPhiMax) {
+      fracPhiMin =
+          fabs(phi(iPhiMin) + 0.5 * m_deltaPhiTower - phiCellMin) / (2 * M_PI / (double)m_segmentation->phiBins());
+      fracPhiMax =
+          fabs(phiCellMax - phi(iPhiMax) + 0.5 * m_deltaPhiTower) / (2 * M_PI / (double)m_segmentation->phiBins());
+      if (fracPhiMin > 1 || fracPhiMax > 1) {
+        warning() << "Fraction if cekk in the tower in phi > 1 not possible!!! " << fracPhiMin << "  " << fracPhiMax
```
jlingema: typos: if -> of, cekk ->cell; add what numbers you print to make the warning more understandable?

Reconstruction/RecCalorimeter/src/components/SingleCaloTowerTool.cpp
```diff
@@ -41,96 +43,134 @@ StatusCode SingleCaloTowerTool::initialize() {
...
   // Loop over a collection of calorimeter cells and build calo towers
-  int iPhi = 0, iEta = 0;
-  float etaCell = 0;
+  // borders of the cell in eta/phi
+  float etaCellMin, etaCellMax;
+  float phiCellMin, phiCellMax;
+  // tower index of the borders of the cell
+  int iPhiMin = 0, iPhiMax = 0;
+  int iEtaMin = 0, iEtaMax = 0;
+  // fraction of cell in eta/phi belonging a tower
```
zaborowska: I would add an extended comment what happens if cell is smaller than tower (all fractions are equal to 1 because first==middle==last tower ?) and what if it is larger (sum of fractions == 1). Maybe also that fractions are calculated based on the size?

Reconstruction/RecCalorimeter/src/components/SingleCaloTowerTool.cpp
```diff
@@ -41,96 +43,134 @@ StatusCode SingleCaloTowerTool::initialize() {
...
+  float ratioEta = 1.0, ratioPhi = 1.0;
+  float fracEtaMin = 1.0, fracEtaMax = 1.0, fracEtaRest = 1.0;
```
zaborowska: I think using Middle/Mid instead of Rest would be more understandable.

faltovaj:
Many thanks for the comments! All the comments should be addressed in the new version. Anna & I spotted a problem in the energy sharing  between clusters (wrong window size was used), also phi edges were not taken correctly. All should be fixed now.
Let me know if there are any further comments,
Jana

faltovaj:
Hi Joschka, I ran again on the changed *.cpp files. I hope it is better now :) 
Thanks, Jana

PR #203, opened 2017-05-15
Combinatorial seeding

### LINKED ISSUE #202: [ACTS-Integration] Add combinatorial seeding
The seeding  currently implemented in ACTS does not use the tracking geometry and therefore isn't blocked by the problems with gcc 6.2, can be added rightaway.

vvolkl:
Fixes #202 .

- Add interface for TrackSeeding
- Integrate the Acts BarrelSeedFinder

Note that this temporarily adds the old ACTS installation to the cmake path so it can be tested. Should be substituted for ACTS 0.04.00 when available.

### DIFF
--- a/Reconstruction/RecTracker/src/components/CombinatorialSeedingTest.cpp
+++ b/Reconstruction/RecTracker/src/components/CombinatorialSeedingTest.cpp
@@ -27,8 +27,6 @@ CombinatorialSeedingTest::CombinatorialSeedingTest(const std::string& name, ISvc
   declareProperty("TrackSeedingTool", m_trackSeedingTool);
 }
 
-CombinatorialSeedingTest::~CombinatorialSeedingTest() {}
-
 StatusCode CombinatorialSeedingTest::initialize() {
   info() << "initialize" << endmsg;
 
@@ -41,7 +39,8 @@ StatusCode CombinatorialSeedingTest::execute() {
   const fcc::PositionedTrackHitCollection* hits = m_positionedTrackHits.get();
   auto seedmap = m_trackSeedingTool->findSeeds(hits);
 
-  for (std::multimap<unsigned int, unsigned int>::iterator it = seedmap.begin(); it != seedmap.end(); ++it) {
+  for (auto seedIdPair: seedmap) {
+    debug() << "trackseed: " << seedIdPair.first << "\t" << seedIdPair.second << endmsg;
   }
 
   return StatusCode::SUCCESS;
--- a/Reconstruction/RecTracker/src/components/CombinatorialSeedingTest.cpp
+++ b/Reconstruction/RecTracker/src/components/CombinatorialSeedingTest.cpp
@@ -1,0 +1,53 @@
+
+#include "DetInterface/IGeoSvc.h"
+
+#include "datamodel/PositionedTrackHitCollection.h"
+#include "datamodel/TrackHitCollection.h"
+#include "datamodel/TrackHitCollection.h"
+
+#include "DD4hep/LCDD.h"
+#include "DD4hep/Volumes.h"
+#include "DDRec/API/IDDecoder.h"
+#include "DDSegmentation/BitField64.h"
+#include "DDSegmentation/CartesianGridXZ.h"
+
+#include <cmath>
+#include <random>
+
+#include "CombinatorialSeedingTest.h"
+#include "RecInterface/ITrackSeedingTool.h"
+#include "RecTracker/TrackingUtils.h"
+
+DECLARE_ALGORITHM_FACTORY(CombinatorialSeedingTest)
+
+CombinatorialSeedingTest::CombinatorialSeedingTest(const std::string& name, ISvcLocator* svcLoc)
+    : GaudiAlgorithm(name, svcLoc) {
+
+  declareProperty("positionedTrackHits", m_positionedTrackHits, "Tracker hits (Input)");
+  declareProperty("TrackSeedingTool", m_trackSeedingTool);
+}
+
+CombinatorialSeedingTest::~CombinatorialSeedingTest() {}
+
+StatusCode CombinatorialSeedingTest::initialize() {
+  info() << "initialize" << endmsg;
+
+  return StatusCode::SUCCESS;
+}
+
+StatusCode CombinatorialSeedingTest::execute() {
+
+  // get hits from event store
+  const fcc::PositionedTrackHitCollection* hits = m_positionedTrackHits.get();
+  auto seedmap = m_trackSeedingTool->findSeeds(hits);
+
+  for (std::multimap<unsigned int, unsigned int>::iterator it = seedmap.begin(); it != seedmap.end(); ++it) {
+  }
+
+  return StatusCode::SUCCESS;
+}
+
+StatusCode CombinatorialSeedingTest::finalize() {
+  StatusCode sc = GaudiAlgorithm::finalize();
+  return sc;
+}
--- a/Reconstruction/RecTracker/src/components/CombinatorialSeedingTest.h
+++ b/Reconstruction/RecTracker/src/components/CombinatorialSeedingTest.h
@@ -17,23 +17,29 @@ class TrackHitCollection;
 class PositionedTrackHitCollection;
 }
 
+
+/*** @class CombinatorialSeedingTest 
+ *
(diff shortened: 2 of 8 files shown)


vvolkl:
Added a [WIP] tag to ensure the changes to init.sh remain temporary and the installations are not mixed.

init.sh
```diff
@@ -3,3 +3,6 @@
 export FCCSWBASEDIR="$( cd "$( dirname "${BASH_SOURCE[0]}" )" && pwd )"
 source /cvmfs/fcc.cern.ch/sw/0.8.1/init_fcc_stack.sh $1
 
+# TODO: temporary, until new ACTS tag available & tests work with gcc6.2
+export CMAKE_PREFIX_PATH=$CMAKE_PREFIX_PATH:/cvmfs/fcc.cern.ch/sw/0.8pre/acts/0.3.0/x86_64-slc6-gcc49-opt/share/cmake/ACTS/
```
jlingema: This mixes acts that is compiled against LCG_83 with the rest of the stack that is compiled against LCG_88. The DD4hep installations are not compatible and this will lead to an inconsistent environment. Let me know if you'd like to have a `acts 0.3.0` install in 0.8.1 (for both gcc62 and gcc49)...

vvolkl: Since the seeding is the only thing that works with 0.3.0 and gcc49, maybe it is better to wait for 0.04.00.

Reconstruction/RecTracker/src/components/CombinatorialSeedingTest.cpp
```diff
@@ -0,0 +1,53 @@
...
+  return StatusCode::SUCCESS;
+}
+
+StatusCode CombinatorialSeedingTest::execute() {
+
+  // get hits from event store
+  const fcc::PositionedTrackHitCollection* hits = m_positionedTrackHits.get();
+  auto seedmap = m_trackSeedingTool->findSeeds(hits);
+
+  for (std::multimap<unsigned int, unsigned int>::iterator it = seedmap.begin(); it != seedmap.end(); ++it) {
```
jlingema: maybe a good opportunity to use `auto` or range-based loop?

Reconstruction/RecTracker/src/components/CombinatorialSeedingTool.h
```diff
@@ -0,0 +1,50 @@
...
+  ~CombinatorialSeedingTool() = default;
+  virtual StatusCode initialize() final;
+  virtual StatusCode finalize() final;
+
+  virtual std::multimap<unsigned int, unsigned int> findSeeds(const fcc::PositionedTrackHitCollection* theHits) final;
+  void createBarrelSpacePoints(Acts::Seeding::BarrelSpacePoints<size_t>& thePoints,
+                               const fcc::PositionedTrackHitCollection* theHits, std::pair<int, int> sIndex);
+
+private:
+  SmartIF<IGeoSvc> m_geoSvc;
```
jlingema: Why not use a `ServiceHandle` here?

Reconstruction/RecTracker/src/components/CombinatorialSeedingTest.h
```diff
@@ -0,0 +1,40 @@
...
+public:
+  CombinatorialSeedingTest(const std::string& name, ISvcLocator* svcLoc);
+
+  ~CombinatorialSeedingTest();
+
+  StatusCode initialize();
+
+  StatusCode execute();
+
+  StatusCode finalize();
```
jlingema: Can you add `override final` here and remove the trivial dtor implementation adding just `= default` here?

Reconstruction/RecTracker/src/components/CombinatorialSeedingTool.cpp
```diff
@@ -0,0 +1,96 @@
...
+  createBarrelSpacePoints(pointsLayer2, theHits, m_seedingLayerIndices1);
+  createBarrelSpacePoints(pointsLayer3, theHits, m_seedingLayerIndices2);
+
+  findHelixSeeds(seedFinderCfg, pointsLayer1, pointsLayer2, pointsLayer3, seeds);
+
+  debug() << "found " << seeds.size() << " track seeds" << endmsg;
+
+  unsigned int trackCounter = 0;
+  for (const auto& seed : seeds) {
+    std::array<float, 15> edmCov;
```
jlingema: Is the covariance matrix empty by design?

Reconstruction/RecTracker/src/components/CombinatorialSeedingTool.h
```diff
@@ -0,0 +1,50 @@
...
+class BitField64;
+}
+}
+
+class CombinatorialSeedingTool : public GaudiTool, virtual public ITrackSeedingTool {
+public:
+  CombinatorialSeedingTool(const std::string& type, const std::string& name, const IInterface* parent);
+  ~CombinatorialSeedingTool() = default;
+  virtual StatusCode initialize() final;
+  virtual StatusCode finalize() final;
```
jlingema: Please add override as well (see PR #209)

Reconstruction/RecTracker/src/components/CombinatorialSeedingTool.h
```diff
@@ -0,0 +1,50 @@
...
+private:
+  SmartIF<IGeoSvc> m_geoSvc;
+  DD4hep::DDSegmentation::BitField64* m_decoder;
+  Gaudi::Property<std::pair<int, int>> m_seedingLayerIndices0{this, "seedingLayerIndices0", {0, 0}};
+  Gaudi::Property<std::pair<int, int>> m_seedingLayerIndices1{this, "seedingLayerIndices1", {0, 1}};
+  Gaudi::Property<std::pair<int, int>> m_seedingLayerIndices2{this, "seedingLayerIndices2", {0, 2}};
+  Gaudi::Property<std::string> m_readoutName{this, "readoutName", "TrackerBarrelReadout"};
+  Gaudi::Property<float> m_rangePhi1{this, "rangePhi1", 0.};
+  Gaudi::Property<float> m_rangePhi2{this, "rangePhi2", 0.5};
+  Gaudi::Property<float> m_dTheta{this, "maxDeltaTheta", 0.5};
```
jlingema: All these properties could use some documentation, especially the `seedingLayerIndices`.

Reconstruction/RecTracker/src/components/CombinatorialSeedingTool.cpp
vvolkl: Yes, it isn't provided at this stage. Maybe we could add a `TrackSeed` to the edm, but I guess they are rarely made persistent.

vvolkl:
Thank you, Joschka, for the comments. I tried to address them in the last commit.

Reconstruction/RecTracker/src/components/CombinatorialSeedingTool.h
```diff
@@ -0,0 +1,58 @@
...
+  /// system and layer ids for the outer barrel layer to be used for seeding
+  Gaudi::Property<std::pair<int, int>> m_seedingLayerIndices2{this, "seedingLayerIndices2", {0, 2}};
+  /// readout used for the barrel seeding layers
+  Gaudi::Property<std::string> m_readoutName{this, "readoutName", "TrackerBarrelReadout"};
+  /// phi window for geometric cut on the middle layer
+  Gaudi::Property<float> m_rangePhi1{this, "rangePhi1", 0.};
+  /// phi window for geometric cut on the outer layer
+  Gaudi::Property<float> m_rangePhi2{this, "rangePhi2", 0.5};
+  /// theta window for geometric cut on middle and outer layer
+  Gaudi::Property<float> m_dTheta{this, "maxDeltaTheta", 0.5};
```
jlingema: Maybe these property names should be synchronised (range vs delta, max vs no prefix). For the `seedingLayerIndices` that means that first is system ID and second is layer id?

vvolkl: I fear that might leed to confusion, as that's what these properties are called in ACTS. As for the `seedingLayerIndices`, that's right.

jlingema: OK, maybe worth mentioning to them then and keep this synchronised, should it change in ACTS.

PR #205, opened 2017-05-16
Fixing segmentation violation in tests

### LINKED ISSUE #204: Segmentation violations when compiled with gcc6.2
Some tests fail with a segmentation violation if the stack + FCCSW is compiled with gcc6.2. Looks like this could be due to some uninitialized members of the `DataHandle` / `PodioDataSvc`.

jlingema:
Fixes #204.

- Initialising all member variables in `PodioDataSvc` and `DataHandle` (we seem to have different optimisation flags for 6.2 / 4.9, will investigate)
- Removed an unused `DataHandle` from the `PythiaInterface`
- Reduced number of events in the configurations used to test `RecCalorimeter` to speed up our tests

To check the changes I enabled gcc6.2 tests in Jenkins.

### DIFF
--- a/Generation/src/components/PythiaInterface.cpp
+++ b/Generation/src/components/PythiaInterface.cpp
@@ -39,7 +39,6 @@ PythiaInterface::PythiaInterface(const std::string& type, const std::string& nam
 
   declareProperty("VertexSmearingTool", m_vertexSmearingTool);
   declarePrivateTool(m_vertexSmearingTool, "FlatSmearVertex/VertexSmearingTool");
-  declareProperty("hepmc", m_hepmchandle, "The HepMC event (output)");
 }
 
 StatusCode PythiaInterface::initialize() {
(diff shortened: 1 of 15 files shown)


vvolkl:
Thanks Joschka, this is a major step towards 0.9! Looks good to me.

Generation/src/components/PythiaInterface.cpp
```diff
@@ -39,7 +39,6 @@ PythiaInterface::PythiaInterface(const std::string& type, const std::string& nam
 
   declareProperty("VertexSmearingTool", m_vertexSmearingTool);
   declarePrivateTool(m_vertexSmearingTool, "FlatSmearVertex/VertexSmearingTool");
-  declareProperty("hepmc", m_hepmchandle, "The HepMC event (output)");
```
vvolkl: This seems to have slipped through during the restructuring of the Generation package. Good catch!

PR #207, opened 2017-05-19
Add Gaussian Vertex distribution

### LINKED ISSUE #206: [Generation] Add Gaussian Vertex Smearing
An additional implementation of the VertexSmearingTool with a normal distribution is needed.

vvolkl:
Fixes #206.

- Adds an implementation of the VertexSmearingTool with a normal distribution
- Synchronizes some parameters with TkLayout Studies


Examples/options/simple_pythia.py
```diff
@@ -16,15 +16,24 @@
...
+from Configurables import ConstPileUp, HepMCFileReader, GaussSmearVertex
+
+from FCCPileupScenarios import FCCPhase1PileupConf as pileupconf
+smeartool = GaussSmearVertex(
+     xVertexMin=pileupconf['xVertexMin'],
+     xVertexMax=pileupconf['xVertexMax'],
+     yVertexMin=pileupconf['yVertexMin'],
+     yVertexMax=pileupconf['yVertexMax'],
+     zVertexMin=pileupconf['zVertexMin'],
+     zVertexMax=pileupconf['zVertexMax'])
```
jlingema: For a Gaussian I'd expect mean and spread / variance here, not min / max?

Generation/src/components/GaussSmearVertex.cpp
```diff
@@ -0,0 +1,94 @@
...
+DECLARE_TOOL_FACTORY(GaussSmearVertex)
+
+/// Standard constructor, initializes variables
+GaussSmearVertex::GaussSmearVertex(const std::string& type, const std::string& name, const IInterface* parent)
+    : GaudiTool(type, name, parent) {
+  declareInterface<IVertexSmearingTool>(this);
+
+  declareProperty("xVertexSigma", m_xsig = 0.0 * Gaudi::Units::mm);
+  declareProperty("yVertexSigma", m_ysig = 0.0 * Gaudi::Units::mm);
+  declareProperty("zVertexSigma", m_zsig = 0.0 * Gaudi::Units::mm);
```
jlingema: Why are these also declared as Properties if they're computed below from the other properties? Also move them into the header.

vvolkl:
Thanks Joschka. I Changed the tool properties to avoid confusion.

Generation/src/components/GaussSmearVertex.h
```diff
@@ -0,0 +1,49 @@
...
+
+  Gaudi::Property<double> m_xmean{this, "xVertexMean", 0.0 * Gaudi::Units::mm, "Mean of x coordinate"};
+  Gaudi::Property<double> m_ymean{this, "yVertexMean", 0.0 * Gaudi::Units::mm, "Mean of y coordinate"};
+  Gaudi::Property<double> m_zmean{this, "zVertexMean", 0.0 * Gaudi::Units::mm, "Mean of z coordinate"};
+  Gaudi::Property<double> m_tmean{this, "tVertexMean", 0.0 * Gaudi::Units::mm, "Mean of t coordinate"};
+
+  /// Direction of the beam to take into account TOF vs nominal IP8, can have
+  /// only values -1 or 1, or 0 to switch off the TOF and set time of
+  /// interaction to zero (default = 1, as for beam 1)
+  Gaudi::Property<int> m_zDir{"beamDirection", 0, "Direction of the beam, possible values: -1, 1 or 0"};
```
jlingema: You say the default is 1 but you set it to 0 below.

Generation/src/components/GaussSmearVertex.cpp
```diff
@@ -0,0 +1,79 @@
...
+
+  if (!sc.isSuccess()) return Error("Could not initialize normal random number generator");
+
+  release(randSvc);
+  return sc;
+}
+
+/// Smearing function
+StatusCode GaussSmearVertex::smearVertex(HepMC::GenEvent& theEvent) {
+  double dx, dy, dz, dt;
```
jlingema: Please initialize variables where you declare them. I.e. also declare them below?

Generation/src/components/GaussSmearVertex.cpp
```diff
@@ -0,0 +1,79 @@
...
+
+  IRndmGenSvc* randSvc = svc<IRndmGenSvc>("RndmGenSvc", true);
+
+
+
+  sc = m_gaussDist.initialize(randSvc, Rndm::Gauss(0., 1));
+  if (sc.isFailure()) return sc;
+
+  std::string infoMsg = " applying TOF of interaction with ";
+  if (m_zDir == -1) {
```
jlingema: I don't see where this is used in the code except for these messages? Is the member necessary?

Generation/src/components/GaussSmearVertex.h
vvolkl: As you mentioned, this property isn't actually used, so I just removed it. I did fix it in the original code, though (FlatSmearVertex).

Generation/src/components/GaussSmearVertex.cpp
vvolkl: Removed it...

Generation/src/components/GaussSmearVertex.cpp
vvolkl: Done!

PR #209, opened 2017-05-22
Fix warnings

jlingema:
With gcc6.2 we have several warnings popping up where externals are using deprecated STL functionality. Furthermore, warnings exist that help making use of C++1y features that couldn't be used because they lead to many such warnings from the externals. This PR implements a workaround.

- Update build flags to the latest version in Gaudi (still need to override to -Wno-shadow due to many warnings from Gaudi itself)
- Include headers from externals as `SYSTEM` (the compiler does not parse them for warnings, errors, etc)
- Add override statements throughout FCCSW where appropriate
- Switch to ninja as default build tool


jlingema:
This should only be merged after a new version of PODIO / fcc-edm was tagged and installed. Otherwise users will be swamped with override warnings from the EDM.

PR #214, opened 2017-06-13
Pileup-Merge algorithm

vvolkl:

- Add components 

To be done:

- [x] Add options file showing usage
- [x] ~~Optionally merge signal and pileup to single collection~~
- [x] Set relations for particles / vertices as in `GenMerge` 
- [x] Chain input signal files: https://github.com/HEP-FCC/podio/issues/3


faltovaj:
Hi Valentin,
thanks for doing this. I have a couple of more general comments. 
1/ Could you please add more comments in the scripts? E.g. write at the beginning of the jobOption file what is is supposed to do.
2/ Do you have an example macro using the PileupMerge algorithm? I would recommend to add a jobOption which will use the min_bias_pool.root and produce the merged pileup file. This file can be further used in the overlay.py. Maybe also the names of the scripts could be a bit longer (e.g. overlaySignalAndPileup.py).
3/ What about a documentation? Do we have a page about the way how pile-up is treated in FCCSW and is it up-to-date?
Thanks a lot,
Jana

Examples/options/overlay.py
```diff
@@ -0,0 +1,79 @@
...
+# next, create the G4 algorithm, giving the list of names of tools ("XX/YY")
+geantsim = G4SimAlg("G4SimAlg",
+                     outputs= ["G4SaveTrackerHits/saveTrackerHits" ])
+geantsim.DataInputs.genParticles.Path="allGenParticles"
+
+from Configurables import PodioInput, PodioMerge
+podin = PodioInput()
+podin.DataOutputs.trackHits.Path = "fileHits"
+podin.DataOutputs.trackClusters.Path = "fileClusters"
+podin.DataOutputs.trackHitsClusters.Path = "fileHitsClusters"
```
faltovaj: I have to say I'm lost here with the inputs and outputs. What are "fileHits", "fileClusters"? Maybe add a comment about what is going on here

FWCore/src/components/PileupMerge.cpp
```diff
@@ -0,0 +1,117 @@
...
+    if (posCaloHitsPresent) {
+      for (auto elem: (*posCaloHitCollection)) {
+        posCaloHitCollOut->push_back(elem.clone());
+      }
+    }
+
+
+    m_store.clear();
+
+  m_trackHitsOut.put(hitCollOut);
```
faltovaj: We can do this for all the collections, right? Or add a switch so that an user can say what collections he is interested in.

FWCore/src/components/PileupMerge.cpp
```diff
@@ -0,0 +1,117 @@
...
+
+  const unsigned int numPileUp = m_pileUpTool->numberOfPileUp();
+  for(unsigned iev = 0; iev < numPileup; ++iev) {
+
+    if ( m_randomizePileup == true ) {
+      if (m_flatDist() < 0.3) { // skip one in three events
+        ++currentEvent;
+      }
+    }
+    currentEvent = (currentEvent + 1) % nEvents; // start over from beginning if necessary
```
faltovaj: Why do you add +1 and not +numPileUp to the currentEvent? I think we will prefer to use different events if possible.

FWCore/src/components/PileupMerge.cpp
```diff
@@ -0,0 +1,117 @@
...
+  fcc::GenVertexCollection* genVertCollOut = new fcc::GenVertexCollection();
+
+  const fcc::CaloHitCollection* caloHitCollection;
+  fcc::CaloHitCollection* caloHitCollOut = new fcc::CaloHitCollection();
+
+  const fcc::PositionedCaloHitCollection* posCaloHitCollection;
+  fcc::PositionedCaloHitCollection* posCaloHitCollOut = new fcc::PositionedCaloHitCollection();
+
+  const unsigned int numPileUp = m_pileUpTool->numberOfPileUp();
+  for(unsigned iev = 0; iev < numPileup; ++iev) {
```
faltovaj: There is a misprint: numPileup -> numPileUp

vvolkl:
Hi Jana,

thanks for the feedback, all valid points! I apologize, you're completely right, the `overlay.py` file is outdated and slipped through, I'll add a more thorough example (and also include it in the test)

FWCore/src/components/PileupMerge.cpp
vvolkl: Hmmm, I think the behavior is correct, but the variable names / documentation are a bit misleading here. For every signal event (call to `execute()`) this code is called `numPileUp` times. `currentEvent` should probably be called `minBiasEventPoolIndex` as it determines which event is read next from the min. bias event pool. I'll update that!

FWCore/src/components/PileupMerge.cpp
vvolkl: Sure! It was a copy-paste-error. But I guess my error is also a sign that the duplications make this hard to maintain and difficult to configure. I can think of  only one better way of implementing a switch like this and letting the user specify the collictions, however, and that is is putting the duplicated code above in tools, one for each type, and then looping over a vector of those tools. I'll try to implement it quickly to make clear what I mean.

FWCore/src/components/PileupMerge.cpp
faltovaj: What I meant is the following:
1st signal event - pileup events number 0 to numPileUp are added
2nd signal event - pileup events number 1 to numPileUp+1 are added
etc.
The pileup in the 1st and 2nd event will be very correlated. This is not what we really want I guess.

FWCore/src/components/PileupMerge.cpp
faltovaj: I see that it is not so that easy :) No need for a quick solution.

FWCore/src/components/PileupMerge.cpp
jlingema: But `currentEvent` is already incremented in the loop above. If you add `numPileUp` here you'll read `2 * numPileUp` events per signal event.

faltovaj: Sorry, my fault. I can see it now.

jlingema: Actually, I just realised that there is another logical error: All the code below the for-loop should go inside...

vvolkl: Oh, Joschka seems to have found the source of the confusion: cloning the particles / hits below must of course happen inside the loop. And `currentEvent` should of course be `m_currentEvent`, to reflect it is a member. A test would have caught this, sorry for not providing one right away! But I guess it is worthwhile to first make the code a little less repetitive by introducing tools.

FWCore/src/components/PileupOverlayAlg.h
```diff
@@ -0,0 +1,69 @@
...
+/*** @class PileupOverlayAlg
+ *  Algorithm for Pileup I/O and merging
+ *
+ * The main purpose of this class is to keep a list of merge tools corresponding to different
+ * datatypes, open a file and let the tools read pileup data and merge it with the signal.
+ * It is thus flexible with regards to the data collection types, as everything is delegated to
+ * tools with a very general interface. One thing this algorithm does do is keep track of the
+ * number of pileup events (via the pileup tool) and the current position in the minimum bias pool
+ * ( which can be randomized ).
+ *
```
vvolkl: I think we had a discussion that this information is already in the commits, so I would just leave the author tag out except when the commits are lost.

FWCore/src/components/PileupTrackHitsMergeTool.cpp
```diff
@@ -0,0 +1,89 @@
+#include "PileupTrackHitsMergeTool.h"
```
vvolkl: Ah yes, I didn't realize the possibility of doing it with tools like this. Seems much better, thanks!

FWCore/FWCore/IEDMMergeTool.h
```diff
@@ -0,0 +1,22 @@
+#ifndef FWCORE_IEDMMERGETOOL_H
+#define FWCORE_IEDMMERGETOOL_H
+
+#include "GaudiKernel/IAlgTool.h"
+
+namespace podio {
+class EventStore;
+}
+
```
vvolkl: Done

FWCore/src/components/PileupParticlesMergeTool.h
```diff
@@ -0,0 +1,45 @@
...
+
+// FCCSW
+#include "FWCore/DataHandle.h"
+#include "FWCore/IEDMMergeTool.h"
+
+namespace fcc {
+class MCParticleCollection;
+class GenVertexCollection;
+}
+
```
vvolkl: Done

Test/TestFWCore/src/components/CreateExampleEventData.h
```diff
@@ -0,0 +1,56 @@
...
+#define TESTFWCORE_CREATEEXAMPLEEVENTDATA
+
+// GAUDI
+#include "GaudiAlg/GaudiAlgorithm.h"
+#include "GaudiKernel/Property.h"
+
+// FCCSW
+#include "FWCore/DataHandle.h"
+
+
```
zaborowska: Missing doxy comment. Maybe say what is tested in here.

Examples/options/overlaySignalAndPileup.py
```diff
@@ -0,0 +1,68 @@
+
+from Gaudi.Configuration import *
+
+
```
zaborowska: Can you add few more comments to that file? What if signalCollections, what happens to them, what is merged etc? This is one of the examples so it may be good to add more documentation.

vvolkl:
retest this please

clementhelsens:
Thanks for this last fix, I'll give it a try and merge the request later

clementhelsens:
This is working as expected now, I'll merge the PR

PR #217, opened 2017-06-19
Adding clang-format pre-commit hook

jlingema:

- Add a simple python script that calls `git clang-format --diff`:
  - `git clang-format --diff` can be used to check the format of staged changes according to .clang-format
  - If mis-matches are found it prints the diff and tells the user to run `git clang-format`
  - `git clang-format` then actually does the re-format (see git clang-format -h after setting up the env for details). Opted for this less intrusive way to give the user a bit of control when things are actually changed
- Set up pre-commit hook in the init script by creating a symlink to the above script to `FCCSW/.git/hooks/pre-commit`


vvolkl:
Superseded by #318

PR #231, opened 2017-08-07
Different thicknesses of layers for inclined ECal

zaborowska:

- allowing different thicknesses of layers in inclined ECal

To be done:

- [x] update xml file with baseline thicknesses: 1x2cm + 7x9cm


hegner:
Some more documentation about it would be needed.

PR #232, opened 2017-08-07
[Acts-Integration ] Use Gaudi Logger in ACTS methods

### LINKED ISSUE #164: [ACTS-Integration] Implement Logger for ACTS
Some workarounds in the ACTS integration PR need to be fixed in the future. The verbosity levels of the Logger should be set from Gaudi, ~~and the B-field should be taken from a central place~~. (Moved to new Issue)

vvolkl:
Fixes #164  .

- Add `Acts-Logger` library that overrides the internally used `Acts::getDefaultLogger` instance.


vvolkl:
Thank you, Alice, for your detailed comments and sorry all the troubles you are having with this logger! Looking back I think this is by no means a model for logging to Gaudi from an external library and I hope we find a different solution. In my eyes this design creates a lot of friction with the gaudi infrastructure - creating a message service for each time "createDefaultLogger" is called, and doing all that from LD_PRELOAD doesn't seem very clean to me. @hegner, do you maybe have some ideas?  So @alicerobson, you were exactly right that the messageLevel didn't arrive in the `FilterPolicy` correctly. I am having a hard time understanding the workings of the MessageSvc and MsgStream as well, so don't ask me why it didn't work, but I finally managed to get the ouputLevel directly from the messageSvc. 

So I tried to implement your suggestions - there is a test in `TestReconstruction` that should print out something like

```
TestACTSLogger      DEBUG Property ['Name': Value] =  'ExtraInputs':[]
                     INFO TestLoggerInfo	TESTING INFO LOGGING LEVEL
                     INFO TestLoggerVerbose	TESTING INFO LOGGING LEVEL
                  VERBOSE TestLoggerVerbose	TESTING DEBUG LOGGING LEVEL
                    DEBUG TestLoggerVerbose	TESTING VERBOSE LOGGING LEVEL
AlgExecStateSvc     DEBUG Service base class initialized successfully
```
which is just what I expect from 
```
  {
  auto testLoggerInfo
        = Acts::getDefaultLogger("TestLoggerInfo", Acts::Logging::INFO);

          ACTS_LOCAL_LOGGER(testLoggerInfo);
          ACTS_INFO("TESTING INFO LOGGING LEVEL");
          ACTS_VERBOSE("TESTING DEBUG LOGGING LEVEL");
          ACTS_DEBUG("TESTING VERBOSE LOGGING LEVEL");
  }

  {

  auto testLoggerVerbose
        = Acts::getDefaultLogger("TestLoggerVerbose", Acts::Logging::VERBOSE);

          ACTS_LOCAL_LOGGER(testLoggerVerbose);
          ACTS_INFO("TESTING INFO LOGGING LEVEL");
          ACTS_VERBOSE("TESTING DEBUG LOGGING LEVEL");
          ACTS_DEBUG("TESTING VERBOSE LOGGING LEVEL");
  }
```
with the correct LD_PRELOAD - otherwise it will still just use the default ACTS Logger.  You might notice that the name of the logger doesn't appear all the way to the left, but I added it manually afterwards, because I just couldn't figure out how the name for the  MsgStream is set. But still,  I would propose to leave it like this until a better solution is found.

vvolkl:
retest this please

vvolkl:
Thanks Alice, done!

Issue #221, opened 2017-06-21
[papas] Enhancement to provide mechanism to use the Gaudi logger in Papas


alicerobson:
Papas should be able to write log messages via the Gaudi logger.

alicerobson:
Now merged

Issue #220, opened 2017-06-19
[papas] Split interface IPapasTool into two 


alicerobson:
The IPapasTool interface should ideally be split  into two parts where one interface is actually writing and the other is actually running something. 
For example, At the moment tools need to implement createOutputStructures but they do nothing in this call.

alicerobson:
Now merged

Issue #218, opened 2017-06-19
[papas] Enhancement: Allow detector to be passed as a service and parameterized


alicerobson:
Enhancement
Papas currently uses CMS detector. To allow more flexibility the detector should be passed as a configurable/parameterized service

alicerobson:
Now merged

PR #234, opened 2017-09-06
Papasdet

alicerobson:
This pull request provides much improved Papas components and allows for the Papas Detector to be provided as a service and for detector parameters to be varied.
It addresses issues 
#220 (interface divided into two)
#221 Gaudi logger now links to Papas ( not the ATS approach used for now as an intermediate solution:- this needs ld_preload and future changes are anticipated)
#218 detector passed as a service and parameterized.

In addition the Clic detector has now been implemented (so users can use Clic or CMS)

This pull request requires papas library  v1.1.0  to be installed.


alicerobson:
Replaced by a newer pull request

PR #238, opened 2017-10-02
Improved Papas including Detector configuration

alicerobson:
This pull request replaces the PapasDet pull request #234. It has been rebased on latest master so that the Pull Request comparison includes the now merged initial papas pull request.
This pull request provides much improved Papas components and allows for the Papas Detector to be provided as a service and for detector parameters to be varied.
It addresses issues
#220 (interface divided into two) actually 3 as both import and export now have separate interface.
#221 Gaudi logger now links to Papas ( not the ATS approach used for now as an intermediate solution:- this needs ld_preload and future changes are anticipated)
#218 detector passed as a service and parameterized.

In addition the Clic detector has now been implemented (so users can use Clic or CMS)
Added linkage information needed for Papas Analyses is also now exported to the root file.

Note that This pull request requires papas library v1.1.0. The init.sh points to a local init_fcc_stack.sh which then makes sure the correct papas library (which is already on the stack) is used. This allows Jenkings tests to be passed. We will need to modify the real init_fcc_stack.sh and remove the local init before merging this pull request.


Sim/SimPapas/SimPapas/IPapasTool.h
```diff
@@ -3,8 +3,14 @@
 
 // Gaudi
 #include "GaudiKernel/IAlgTool.h"
+// Papas
+#define WITHSORT 1
```
cbernet: Can this WITHSORT be replaced by a static variable in the papas namespace? 
There is a risk of collision if another WITHSORT is defined somewhere else by somebody else

Sim/SimPapas/options/simple_papastool_CMS.py
```diff
@@ -0,0 +1,108 @@
...
+from GaudiKernel import SystemOfUnits as units
+
+## read in generated particles from ROOT via podio
+#podioevent   = FCCDataSvc("EventDataSvc", input="./ee_ZH_Zmumu_Hbb.root")
+podioevent   = FCCDataSvc("EventDataSvc", input="./ee_Z_bbbar.root")
+
+from Configurables import PodioInput, ReadTestConsumer
+podioinput = PodioInput("PodioReader", collections=["GenVertex", "GenParticle"], OutputLevel=DEBUG)
+
+from Configurables import CMSFieldSvc, CMSTrackerSvc, CMSEcalSvc, CMSHcalSvc
```
cbernet: Is there a way to put all the papas thing in a specific configuration file that would then be imported in higher level configuration files? 

The user should be able to do: 

import papasalg

possibly change stuff like the detectors

add papasalg to the list of algorithms, as done at the very end of this file.

cbernet: One could have one papas configuration file per detector, 
but even then, there are things in common, and these common things should be organized in a common module somehow

Sim/SimPapas/src/PapasAlg.cpp
```diff
@@ -1,34 +1,58 @@
...
 }
 
 StatusCode PapasAlg::initialize() {
-
   for (auto& toolname : m_toolNames) {
     m_tools.push_back(tool<IPapasTool>(toolname));
   }
+  m_importTool = tool<IPapasImportTool>(m_importToolName);
+  m_exportTool = tool<IPapasExportTool>(m_exportToolName);
+  SmartIF<IPapasDetSvc> papasDetSvc;  // curiously does not work if joined with following line
```
cbernet: Could this be because one is forced to use an explicit constructor? 
I guess that 

    SmartIF<IPapasDetSvc> papasDetSvc(service(m_detServiceName))

should work? but no big deal anyway

Sim/SimPapas/src/PapasAlg.cpp
```diff
@@ -42,25 +66,30 @@ StatusCode PapasAlg::execute() {
...
+    m_exportTool->createOutputStructures();
+    //import to create SimParticles and links between Sim and GenParticles
+    m_importTool->run(pevent, m_importParticleLinks, m_spDetector);
+    // now do the real work
     for (auto tool : m_tools) {
       // run each tool
-      tool->run(pevent);
+      tool->run(pevent, m_spDetector);
     }
+    m_exportTool->run(pevent, m_importParticleLinks);
```
cbernet: this line sounds funny (exporting importing particles?)
what is m_importParticleLinks? 
is the name well chosen?

Sim/SimPapas/src/PapasAlg.h
```diff
@@ -35,16 +41,24 @@ class PapasAlg : public GaudiAlgorithm {
...
+  std::shared_ptr<papas::Detector> m_spDetector;  ///< papas detector
   /// Stores the linkages between papas objects (eg clusters, tracks, particles)
   papas::Nodes m_history;
-  std::vector<IPapasTool*> m_tools;      ///<vector of tools to be run
-  std::vector<std::string> m_toolNames;  ///< names of tools to be run
-  long m_eventno;                        ///< the papas Event number, incremented for each event processed
+  std::vector<IPapasTool*> m_tools;       ///< vector of tools to be run
+  IPapasImportTool* m_importTool;         ///< import tool to be run
+  IPapasExportTool* m_exportTool;         ///< export tool to be run
+  std::vector<std::string> m_toolNames;   ///< names of tools to be run
```
cbernet: why are the names necessary?
for m_tools, couldn't you use a 

```c++
std::vector< std::pair<std::string, IPapasTool*> >
```

instead?

Sim/SimPapas/src/PapasAlg.h
```diff
@@ -35,16 +41,24 @@ class PapasAlg : public GaudiAlgorithm {
...
+  std::string m_importToolName;           ///< name of import tool to be run
+  std::string m_exportToolName;           ///< name of export tool to be run
+  std::unordered_map<papas::Identifier, int> m_importParticleLinks; ///< map to contain links between GenParticles and SimParticles
+  long m_eventno;                         ///< the papas Event number, incremented for each event processed
```
cbernet: Is that really necessary? could we use the gaudi event number?

Sim/SimPapas/src/PapasBuildBlocksTool.cpp
```diff
@@ -21,8 +20,9 @@ StatusCode PapasBuildBlocksTool::clear() {
   return StatusCode::SUCCESS;
 }
 
-StatusCode PapasBuildBlocksTool::run(papas::Event& pevent) {
+StatusCode PapasBuildBlocksTool::run(papas::Event& pevent, std::shared_ptr<papas::Detector> det) {
   auto history = papas::Nodes();
+  (void)det;  // avoid warninf
```
cbernet: Could you please give me more information about this line? why it's there and what it does? 
Also, please improve the comment

Sim/SimPapas/src/PapasExportParticlesTool.cpp
```diff
@@ -1,14 +1,21 @@
...
 
 DECLARE_TOOL_FACTORY(PapasExportParticlesTool)
 
 PapasExportParticlesTool::PapasExportParticlesTool(const std::string& aType, const std::string& aName,
                                                    const IInterface* aParent)
     : GaudiTool(aType, aName, aParent) {
-  declareInterface<IPapasTool>(this);
+  declareInterface<IPapasExportTool>(this);
   declareProperty("recparticles", m_particlesHandle);
+  declareProperty("genparticles", m_iGenpHandle);
```
cbernet: the name is not symmetric with the line just above. why not m_genParticlesHandle?

Sim/SimPapas/src/PapasExportParticlesTool.cpp
```diff
@@ -18,13 +25,25 @@ StatusCode PapasExportParticlesTool::initialize() {
...
   // convert each papas particle into an MCParticle
   std::string ptype = m_particleSubtype;
+  papas::HistoryHelper history(pevent);
+  
+  //put the original gen particles into a vector so they can be used to
+  //make the link association with the reconstructed particle
+  const fcc::MCParticleCollection* ptcs = m_iGenpHandle.get();
+  std::vector<fcc::ConstMCParticle> vecPtcs;
+  for (const auto& p : *ptcs) {
+     vecPtcs.push_back(p);
```
cbernet: but isn't ptcs already a vector? why do we need a second one?

Sim/SimPapas/src/PapasImportParticlesTool.cpp
```diff
@@ -35,31 +36,35 @@ StatusCode PapasImportParticlesTool::clear() {
...
+    auto p4 = vecPtcs[a].p4();
     tlv.SetXYZM(p4.px, p4.py, p4.pz, p4.mass);
-    p4 = b.p4();
+    p4 = vecPtcs[b].p4();
     tlv2.SetXYZM(p4.px, p4.py, p4.pz, p4.mass);
     return tlv.E() > tlv2.E();
   });
-
+  
   // Now construct the papas Particles (providing they are stable)
```
cbernet: need more explicit comment. What kind of papas Particles are constructed? 
I understand that we are talking about the reconstructed particles from the comment,
but the code seems to indicate we are constructing the simulated particles before papas

Sim/SimPapas/src/PapasSimulatorTool.h
```diff
@@ -17,7 +17,7 @@ class MCParticleCollection;
 /** @class PapasSimulatorTool Sim/SimPapas/src/PapasSimulatorTool.h PapasSimulatorTool.h
  *
  *  Runs the papas simulation algorithm. This creates a list of particles from Pythia generated particles.
```
cbernet: which particles are we talking about? 
"list of particles" not explicit enough

Sim/SimPapas/src/PapasSimulatorTool.h
```diff
@@ -17,7 +17,7 @@ class MCParticleCollection;
...
- *  It then creates particle flow simulated and true and smeared clusters and tracks. All of these
+ *  It then creates particle-flow-simulated true and smeared clusters and tracks. All of these
```
cbernet: particle flow does the reconstruction from the tracks and the clusters. 
it does not simulate them, and does not play a role in the creation of true and smeared clusters. 

replace by: 

It then simulates the interaction of these particles in the detector to produce true and smeared clusters and tracks, that will be used in the particle flow reconstruction.

Sim/SimPapasDetector/CMakeLists.txt
```diff
@@ -0,0 +1,22 @@
+################################################################################
```
cbernet: I don't know the software architecture of FCCSW very well, but why is SimPapasDetector not in Sim/SimPapas?

Sim/SimPapasDetector/src/CMSDetectorSvc.cpp
```diff
@@ -0,0 +1,40 @@
...
+DECLARE_SERVICE_FACTORY(CMSDetSvc)
+
+CMSDetSvc::CMSDetSvc(const std::string& name, ISvcLocator* svc) : base_class(name, svc), m_detector(nullptr) {}
+
+StatusCode CMSDetSvc::initialize() {
+  if (Service::initialize().isFailure()) {
+    error() << "Unable to initialize Service()" << endmsg;
+    return StatusCode::FAILURE;
+  }
+  SmartIF<IPapasFieldSvc> papasFieldSvc;
```
cbernet: What is SmartIF?

Sim/SimPapas/src/PapasBuildBlocksTool.cpp
alicerobson: The blockbuilder tol uses the IPapasTool interface and this has an optional argument which is the papasdetector. In the case of the Blockbuilder tool the papas detector " det" is not used. This generates a warning message.  (void)det is a trick to suppress the warning message.

Sim/SimPapas/src/PapasExportParticlesTool.cpp
alicerobson: Yes indeed! thanks.

Sim/SimPapas/src/PapasExportParticlesTool.cpp
alicerobson: trying too hard... Thanks

Sim/SimPapas/src/PapasImportParticlesTool.cpp
alicerobson: Improved

Sim/SimPapas/src/PapasSimulatorTool.h
alicerobson: done

Sim/SimPapas/src/PapasSimulatorTool.h
alicerobson: done

Sim/SimPapasDetector/src/CMSDetectorSvc.cpp
alicerobson: The documentation says:
Small smart pointer class with automatic reference counting for IInterface.

SmartIF simplifies the interaction with components in Gaudi by implementing an automatic reference counting and the casting (via IInterface::queryInterface).

Any the wiser? This is the way other services have been accessed in FCCSW and I copied their approach.

Sim/SimPapas/src/PapasAlg.h
alicerobson: If I understand correctly, its not possible to pass any old argument through from the python configuration file (such as your vector above), but one can pass a list of strings. So the PapasAlg takes a list of tool names and then the individual tools are picked up inside the Algorithm. My approach is based on similar code in other parts of FCCSW

Sim/SimPapas/src/PapasAlg.cpp
alicerobson: No! -  name now changed.

Sim/SimPapasDetector/CMakeLists.txt
alicerobson: I think it would be good to discuss with @benedikt the best place to put the papas files within the FCC structures.

cbernet:
Hi @alicerobson, 

@hegner checked your PR and also finds it very clean and nice. He thinks it can be merged.
I'd like to merge, but I do not see the additional commits that would answer the comments in my review. Could you please check? Thanks!

Sim/SimPapasDetector/src/CMSDetectorSvc.cpp
cbernet: ok, thank you.

Sim/SimPapas/src/PapasBuildBlocksTool.cpp
cbernet: Still, I think the comment should be developped and improved a bit. most people will not know what is "warninf" (or maybe do you mean warning??) if yes, just say write which warning.

Sim/SimPapas/src/PapasAlg.h
cbernet: ok

Sim/SimPapas/options/simple_papastool_CMS.py
alicerobson: I have now managed to do this with a little help from Anna who pointed out an example of Valentin!

cbernet: thanks @zaborowska @vvolkl !

Sim/SimPapas/src/PapasAlg.h
alicerobson: Now done.

Sim/SimPapasDetector/CMakeLists.txt
alicerobson: I have moved as you have suggested

alicerobson:
Replaced by another pull request and now merged

Issue #239, opened 2017-10-09
Allow access to particle trajectory data in Full Sim


vvolkl:
For debugging and visualisation purposes, the full particle trajectories as simulated by geant are sometimes useful. They should be accessible within the framework.

vvolkl:
Closed by #240

PR #240, opened 2017-10-09
Add Tool to save Geant4 trajectory data

vvolkl:
Addresses #239.

Adds a new SaveOutput tool for Geant4 particle trajectory data. There is no type in the edm that corresponds exactly to this kind of data, so I went with TrackerHits, which is related but slightly redundant. The trajectories are then saved together in one collection with the TrackID used to distinguish different trajectories.


vvolkl:
retest this please

PR #243, opened 2017-10-10
HepMC Converter Update - Final state filter made optional

vvolkl:
For some applications it is necessary to not only convert the final, but all particles. To achieve this, the  HepMCConverter must be expanded to keep track of the vertex/particle relations, and an option added to make filtering particles optional (in order not to break existing applications the default is still to convert only particles with `status == 1`).


Generation/src/components/HepMCToEDMConverter.cpp
```diff
@@ -29,43 +32,79 @@ StatusCode HepMCToEDMConverter::execute() {
...
-         particle_i != (*vertex_i)->particles_end(HepMC::children);
-         ++particle_i) {
-      // take only final state particles
-      if ((*particle_i)->status() != 1) continue;
+    /// create production vertex, if it has not already been created and logged in the map
+    HepMC::GenVertex* productionVertex = (*particle_i)->production_vertex();
+    if (nullptr != productionVertex) {
+      if (hepmcToEdmVertexMap.find(productionVertex) != hepmcToEdmVertexMap.end()) {
+        // vertex already in map, no need to create a new one
+        particle.startVertex(hepmcToEdmVertexMap[productionVertex]);
```
gouskos: please change: from "hepmcToEdmVertexMap[productionVertex]" to "hepmcToEdmVertexMap[decayVertex]"

vvolkl: Hi Loukas @gouskos,

thanks for your comments and sorry for the late reply! The way you write it it doesn't work - `decayVertex` isn't declared until line 83 below. There was a copy & paste bug introduced by me, which I already fixed in https://github.com/HEP-FCC/FCCSW/pull/243/commits/76c9252a2fc332d818b4fdfa2ced8e924489b446
it's essentially the same thing but on L87. Could I ask you to check that this doesn't work already? I'll have a look why the tests are failing!
Cheers,
Valentin

vvolkl:
retest this please

vvolkl:
retest this please

vvolkl:
retest this please

clementhelsens:
Thanks Valentin, I merge this PR

PR #236, opened 2017-09-12
change all gaudirun to fccrun

clementhelsens:
replace #235

PR #245, opened 2017-10-27
Calorimeter cell positions for topo-clustering and jet reconstruction

CoralieNeubueser:
1) This is an addition to the volume positions, using the dd4hep segmentation.
Temporarily enables the retrieval of cell positions for all Calorimeter systems and segmentations, an example for the full digitisation chain of the Barrel is given in the runFullCaloSystem_SimAndFullDigitisation.py script.

2) TileCal geometry changed to partially replace Steel with Pb absorbers, discussed and approved within the FCC Calorimeter group. Implications are a better single particle resolution in standalone and combined reconstruction due to reduced non-compensation (e/h=1.1), the sampling fraction changes from 3.2% to 2.7%. The calibration to hadron scale changes from 2.9% to 2.4%.

To be done:
- geometry independent way to extract cell positions and volumes (needs further discussions with dd4hep people)


hegner:
test this please

CoralieNeubueser:
Hello,

I checked the tests output and the issue is some timeout after 1500.01sec. The running scripts are fine, the code is fine.
Can we please allow for longer checking time of the tests?

Cheers,
Coralie

On 1 Nov 2017, at 17:50, fcc-bot <notifications@github.com<mailto:notifications@github.com>> wrote:


Build or tests failed.

hegner:
Actually we should make sure that our test times get a bit reduced. 1500s for a test actually sounds like something is stalled/wrong there. What execution times do yet get locally?

CoralieNeubueser:
I get for that specific test (RecCalorimeter.simulateFullCaloSystemFullDigitisation) 127.38sec
The timeout error from Jenkins I also get for 6 other tests which don’t show any problem when testing locally…
Maybe there is something wrong with Jenkins?!

On 2 Nov 2017, at 10:16, hegner <notifications@github.com<mailto:notifications@github.com>> wrote:


Actually we should make sure that our test times get a bit reduced. 1500s for a test actually sounds like something is stalled/wrong there. What execution times do yet get locally?

hegner:
well, we could make a new test run. If that doesn't help, we need to investigate closer

hegner:
retest this please

Detector/DetComponents/src/CreateCellPositions.cpp
```diff
@@ -0,0 +1,113 @@
...
+            << outGlobal[2] / dd4hep::mm << endmsg;
+
+    // get PhiEta segmentation
+    DD4hep::DDSegmentation::GridPhiEta* segmentation;
+    segmentation = dynamic_cast<DD4hep::DDSegmentation::GridPhiEta*>(
+        m_geoSvc->lcdd()->readout(m_readoutName).segmentation().segmentation());
+    if (segmentation == nullptr) {
+      error() << "There is no phi-eta segmentation!!!!" << endmsg;
+      return StatusCode::FAILURE;
+    }
```
faltovaj: This part (retrieving the segmentation) could be moved in the initialization, not to be called for each event & each hit.

Detector/DetComponents/src/CreateCellPositions.h
```diff
@@ -0,0 +1,56 @@
...
+  StatusCode execute();
+  /**  Finalize.
+   *   @return status code
+   */
+  StatusCode finalize();
+
+private:
+  /// Pointer to the geometry service
+  ServiceHandle<IGeoSvc> m_geoSvc;
+  /// Name of the electromagnetic calorimeter readout
```
faltovaj: Update the comment, it is not just ecal :)

Reconstruction/RecCalorimeter/tests/options/runFullCaloSystem_SimAndFullDigitisation.py
```diff
@@ -0,0 +1,197 @@
+import os
+import numpy as np
+
+# simulations setup                                                                                                                                                                                                              
+energy=1000000
```
faltovaj: Why do you run for such high energy? Is it that you want to have signal in most of the detector in one direction? Maybe the energy could be lowered for the testing.

Reconstruction/RecCalorimeter/tests/options/runFullCaloSystem_SimAndFullDigitisation.py
```diff
@@ -0,0 +1,197 @@
...
+num_events=1
```
clementhelsens: for the events, it can be in fact given as argument
if args.nevents is not None:
    num_events = args.nevents
and then when you call the script, you add the option --nevents=1234, if nothing, the default is used

CoralieNeubueser: Hi Clement, thanks for the suggestion! But could you please be more specific? Do i need to import the GaudiKernel.ProcessJobOptions ? Or something else?
Thanks,
Coralie

Reconstruction/RecCalorimeter/tests/options/runFullCaloSystem_SimAndFullDigitisation.py
CoralieNeubueser: I kept it high to make sure I have deposits in all sub detectors that I have tested.. but I can lower it now.

Reconstruction/RecCalorimeter/tests/options/runFullCaloSystem_SimAndFullDigitisation.py
clementhelsens: sorry for not mentioning it before, yes, you need to import something before, so the full syntax would be:

from FWCore.joboptions import parse_standard_job_options
args = parse_standard_job_options()
num_events=100
if args.nevents is not None:
    num_events = args.nevents

CoralieNeubueser: Thank you, I added it, but it only gives me an error.. it's complaining that gaudirun.py (which I call to run my script) does not know the option --nevents..

clementhelsens: ah yes, it's because you need to use fccrun
#236 was supposed to remove all references to gaudirun, but was not approved yet

zaborowska:
ok to test

zaborowska:
retest this please

CoralieNeubueser:
The pr will be re-opened. The cell positions can be retrieved in a more general way after a few changes in the calorimeter geometry.

PR #246, opened 2017-10-31
Interactive mode

nalipour:
Addresses # .

- SimG4Svc changed in order to run Geant4 in interactive mode and use the Geant4 command lines in the terminal. It is useful for debugging but especially for the execution of DAWN visualisation.
- An example of script for running FCCSW in interactive mode: Examples/options/interactive_mode.py

### DIFF
--- a/Sim/SimG4Components/src/SimG4Svc.cpp
+++ b/Sim/SimG4Components/src/SimG4Svc.cpp
@@ -66,8 +66,11 @@ StatusCode SimG4Svc::initialize() {
   m_runManager.Initialize();
 
   if (m_interactiveMode) {
-    visManager->Initialize();
-    session->SessionStart();
+    m_visManager = std::make_unique<G4VisExecutive>();
+    m_visManager->Initialize();
+
+    m_session = std::make_unique<G4UIterminal>();
+    m_session->SessionStart();
   }
 
   // Attach user actions
--- a/Sim/SimG4Components/src/SimG4Svc.cpp
+++ b/Sim/SimG4Components/src/SimG4Svc.cpp
@@ -65,12 +65,8 @@ StatusCode SimG4Svc::initialize() {
 
   m_runManager.Initialize();
 
-  std::unique_ptr<G4VisManager> visManager(new G4VisExecutive);
-
   if (m_interactiveMode) {
     visManager->Initialize();
-    // Define UI terminal for interactive mode
-    std::unique_ptr<G4UIsession> session(new G4UIterminal());
     session->SessionStart();
   }
 
--- a/Sim/SimG4Components/src/SimG4Svc.cpp
+++ b/Sim/SimG4Components/src/SimG4Svc.cpp
@@ -8,6 +8,11 @@
 #include "G4UImanager.hh"
 #include "G4VModularPhysicsList.hh"
 
+#include "G4UIsession.hh"
+#include "G4UIterminal.hh"
+#include "G4VisExecutive.hh"
+#include "G4VisManager.hh"
+
 DECLARE_SERVICE_FACTORY(SimG4Svc)
 
 SimG4Svc::SimG4Svc(const std::string& aName, ISvcLocator* aSL) : base_class(aName, aSL) {
@@ -54,12 +59,20 @@ StatusCode SimG4Svc::initialize() {
   m_runManager.SetUserInitialization(m_detectorTool->detectorConstruction());
 
   G4UImanager* UImanager = G4UImanager::GetUIpointer();
-  for (auto command: m_g4PreInitCommands) {
+  for (auto command : m_g4PreInitCommands) {
     UImanager->ApplyCommand(command);
   }
 
-
   m_runManager.Initialize();
+  if (m_interactiveMode) {
+    G4VisManager* visManager = new G4VisExecutive;
+    visManager->Initialize();
+    // Define UI terminal for interactive mode
+    G4UIsession* session = new G4UIterminal;
+    session->SessionStart();
+    delete session;
+  }
+
   // Attach user actions
   m_runManager.SetUserInitialization(m_actionsTool->userActionInitialization());
   // Create regions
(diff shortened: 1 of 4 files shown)


hegner:
test this please

nalipour:
Hi Valentin,

The requested changes are submitted (as you already know :)).

Thanks,
Nilou

vvolkl:
Yes, Thank you! I think there is some small issue with CI, to me this PR looks good.

Sim/SimG4Components/src/SimG4Svc.cpp
```diff
@@ -54,12 +59,20 @@ StatusCode SimG4Svc::initialize() {
...
 
-
   m_runManager.Initialize();
+  if (m_interactiveMode) {
+    G4VisManager* visManager = new G4VisExecutive;
+    visManager->Initialize();
+    // Define UI terminal for interactive mode
+    G4UIsession* session = new G4UIterminal;
+    session->SessionStart();
+    delete session;
```
hegner: can't we work with smart pointers here, instead of bare ones?

Sim/SimG4Components/src/SimG4Svc.cpp
```diff
@@ -54,12 +59,20 @@ StatusCode SimG4Svc::initialize() {
...
   G4UImanager* UImanager = G4UImanager::GetUIpointer();
-  for (auto command: m_g4PreInitCommands) {
+  for (auto command : m_g4PreInitCommands) {
     UImanager->ApplyCommand(command);
   }
 
-
   m_runManager.Initialize();
+  if (m_interactiveMode) {
+    G4VisManager* visManager = new G4VisExecutive;
```
hegner: who's owner of G4VisManager in the end? I don't see a matching delete

Sim/SimG4Components/src/SimG4Svc.cpp
nalipour: Hi Benedikt,

Where should I keep the additional markdown file?
I have changed the bare pointer solution into smart ones. The visManager, should be deleted at the end of the program, otherwise it causes a segmentation fault.

Thanks,
Nilou

Sim/SimG4Components/src/SimG4Svc.cpp
```diff
@@ -64,13 +64,14 @@ StatusCode SimG4Svc::initialize() {
   }
 
   m_runManager.Initialize();
+
+  std::unique_ptr<G4VisManager> visManager(new G4VisExecutive);
```
hegner: if the G4VisManager has to survive until the end of the program it cannot be a unique_ptr in the local scope.

Sim/SimG4Components/src/SimG4Svc.cpp
```diff
@@ -64,13 +64,14 @@ StatusCode SimG4Svc::initialize() {
...
+
   if (m_interactiveMode) {
-    G4VisManager* visManager = new G4VisExecutive;
     visManager->Initialize();
     // Define UI terminal for interactive mode
-    G4UIsession* session = new G4UIterminal;
+    std::unique_ptr<G4UIsession> session(new G4UIterminal());
```
hegner: You could have a look at http://en.cppreference.com/w/cpp/memory/unique_ptr/make_unique

hegner:
The doc can go to the dir `Sim/doc`

vvolkl:
In the docs, could you make explicit also that this is a debugging feature? I have actually no idea what will happen when you try to run "/run/beamOn" (I suspect it will just throw an exception).

Sim/SimG4Components/src/SimG4Svc.cpp
nalipour: For the case that I am using (running DAWN), the pointer survives until the end of the program and there is no segmentation fault. Should I then follow the example of the RunManager?

Sim/SimG4Components/src/SimG4Svc.cpp
nalipour: Is there any problem using unique_ptr in this case ?

vvolkl: Hi Nilou,

no, `make_unique` essentially does the same as the line above, but makes for a clearer syntax ( you never even have to write `new`) and is recommended for a few more technical reasons.

Sim/SimG4Components/src/SimG4Svc.cpp
vvolkl: Exactly, if the lifetime for the Visualisation Manager is the same as for the SimG4Svc, then it should move out of the local scope and be a member of SimG4Svc, so it has ownership.

Sim/SimG4Components/src/SimG4Svc.h
```diff
@@ -77,8 +84,14 @@ class SimG4Svc : public extends1<Service, ISimG4Svc> {
   Gaudi::Property<std::vector<std::string>> m_regionToolNames{
       this, "regions", {}, "Names of the tools that create regions and fast simulation models"};
 
+  Gaudi::Property<bool> m_interactiveMode{this, "InteractiveMode", false, "Enter the interactive mode"};
+
   /// Run Manager
   sim::RunManager m_runManager;
+
+  std::unique_ptr<G4VisManager> visManager = std::make_unique<G4VisExecutive>();
```
vvolkl: One last thing: Could you rename the `visManager` to `m_visManager`, as in the style guide, and only initialize the pointer in `initialize()`? So that would be

`std::unique_ptr<G4VisManager> m_visManager{nullptr};`

and then

`m_visManager = std::make_unique<G4VisExecutive>();`

in the `.cpp`, and analogous for `session`.  Thanks!

nalipour:
Hi Valentin,

Thanks for your comments. The changes have been committed.

Cheers,
Nilou

vvolkl:
retest this please

PR #252, opened 2017-11-21
Update Tracker factories to construct positive and negative endcaps at the same time

vvolkl:
This should speed up the construction of the geometry and help with the readout.


Detector/DetFCChhTrackerTkLayout/src/TkLayoutConicalEndcap_Geo.cpp
```diff
@@ -14,148 +13,171 @@ using DD4hep::Geometry::PlacedVolume;
...
   std::string detName = xmlDet.nameStr();
   DetElement worldDetElement(detName, xmlDet.id());
+  DetElement posEcapDetElement(worldDetElement, "posEcap", 0);
+  DetElement negEcapDetElement(worldDetElement, "negEcap", 1);
 
   Acts::ActsExtension::Config actsBarrelConfig;
-  actsBarrelConfig.isEndcap             = true;
+  actsBarrelConfig.isEndcap = true;
   Acts::ActsExtension* worldDetExt = new Acts::ActsExtension(actsBarrelConfig);
   worldDetElement.addExtension<Acts::IActsExtension>(worldDetExt);
```
jhrdinka: The extension needs to be added to the two endcap DetElements posEcapDetElement and negEcapDetElement

Detector/DetFCChhTrackerTkLayout/compact/FCChh_v3.03_Definitions.xml
```diff
@@ -3813,19 +2555,24 @@
                 </rings>
             </discZPls>
             <discZPls id="2" z="10985.605*mm" zmin="10976.312*mm" zmax="10994.899*mm" rmin="49.500*mm" rmax="1550.163*mm" bigDelta="5.000*mm">
```
jhrdinka: Starting from here there is also a duplication of the IDs. Also I do not fully understand how this works, because there is a change in how the discs are declared starting from here. It is done similar as in the inner or outer endcap but the constructor is different (Conical) and in Acts I do not get any modules for these layers.

Detector/DetFCChhTrackerTkLayout/compact/Tracker.xml
```diff
@@ -10,7 +10,7 @@
         </readout>
         <readout name="TrackerEndcapReadout">
             <segmentation type="CartesianGridXZ" grid_size_x="0.005*mm" grid_size_z="0.010*mm"/>
-            <id>system:5,disc:5,component:18,x:-15,z:-15</id>
+            <id>system:5,posneg:1,disc:5,component:17,x:-15,z:-15</id>
         </readout>
     </readouts>
     <include ref="FCChh_v3.03_Definitions.xml"/>
```
jhrdinka: Now all the detector assemblies should be removed, since they are not needed anymore because the endcaps are grouped together and we can anyway not distinguish Outer and Inner tracker in the readout.

Detector/DetFCChhTrackerTkLayout/compact/FCChh_v3.03_Definitions.xml
```diff
@@ -3532,27 +2295,6 @@
...
-            <discZPls id="6" z="16000.000*mm" zmin="15990.707*mm" zmax="16009.294*mm" rmin="49.500*mm" rmax="1550.163*mm" bigDelta="5.000*mm">
-                <rings repeat="15" smallDelta="2.500*mm" rPhiOverlap="0.500*mm" nRPhiSegments="4" rOverlap="0.500*mm"/>
-            </discZPls>
-        </discs>
-    </detector>
-    <detector name="FwdECAPneg" id="28" type="TkLayoutEcapTracker" readout="TrackerEndcapReadout">
-        <dimensions rmin="49.500*mm" rmax="1550.163*mm" zmin="9990.707*mm" zmax="16009.294*mm" reflect="true"/>
-        <sensitive type="SimpleTrackerSD"/>
-        <discs repeat="12">
             <discZPls id="1" z="10000.000*mm" zmin="9990.707*mm" zmax="10009.293*mm" rmin="49.500*mm" rmax="1550.163*mm" bigDelta="5.000*mm">
```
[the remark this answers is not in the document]
vvolkl: These id's aren't used, but they can be updated in the post-processing of the xml, might be good for readability

Detector/DetFCChhTrackerTkLayout/src/TkLayoutEndcap_Geo.cpp
```diff
@@ -145,17 +145,22 @@ static DD4hep::Geometry::Ref_t createTkLayoutTrackerEndcap(DD4hep::Geometry::LCD
...
-  DD4hep::Geometry::RotationX envelopeNegRotation(envelopeNegRotAngle);
-  PlacedVolume placedEnvelopeVolume = motherVol.placeVolume(envelopeVolume, envelopeNegRotation * envelopeTranslation);
-  placedEnvelopeVolume.addPhysVolID("system", xmlDet.id());
-  worldDetElement.setPlacement(placedEnvelopeVolume);
+
+  DD4hep::Geometry::RotationX envelopeNegRotation(dd4hep::pi);
+  DD4hep::Geometry::RotationX envelopePosRotation(0.);
+  PlacedVolume placedEnvelopeVolume = bothEndcaps.placeVolume(envelopeVolume, envelopePosRotation * envelopeTranslation);
+  PlacedVolume placedNegEnvelopeVolume = bothEndcaps.placeVolume(envelopeVolume, envelopeNegRotation * envelopeTranslation);
+  placedEnvelopeVolume.addPhysVolID("posneg", 0);
```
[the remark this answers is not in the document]
vvolkl: This constructor isn't used anymore, I'll remove it

Detector/DetFCChhTrackerTkLayout/compact/FCChh_v3.03_Definitions.xml
vvolkl: These Id's aren't used anymore, the modules are connected to the detelements that don't exist for repeatedly placed discs. either you get them from the first disc, or I will try cloning the DetElements

Detector/DetFCChhTrackerTkLayout/src/TkLayoutConicalEndcap_Geo.cpp
vvolkl: That's true, I will try to substitute this by a clone of the DetElement, this should copy the Extensions as well.

Detector/DetFCChhTrackerTkLayout/compact/FCChh_v3.03_Definitions.xml
jhrdinka: Why do the DetElements not exist for repeatedly placed discs? Sensitive detector elements need to be unique and should be dd4hep::DetElements.

Detector/DetFCChhTrackerTkLayout/src/TkLayoutEndcap_Geo.cpp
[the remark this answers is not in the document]
jhrdinka: It seems that it is used here: https://github.com/vvolkl/FCCSW/blob/ed5e76bb9c3755d6e7c6f82cf13661af77c9a8be/Detector/DetFCChhTrackerTkLayout/compact/FCChh_v3.03_Definitions.xml#L560

Detector/DetFCChhTrackerTkLayout/src/TkLayoutConicalEndcap_Geo.cpp
jhrdinka: What is exactly the advantage of deep copying the DetElements, from creating them directly? Since the clone method was not tested for some time in DD4hep and we are not sure if the hierachies of the DetElements and their extensions are copied correctly, I am not sure if this is the way to go.

jhrdinka:
I just tested the volumeIDs.  For the ForwardEndcap the component DetElements (which represent our sensitive detector parts) seem all to have the same volumeID. The other endcaps I can not test right now until the update.
 Also when the DD4hep geometry is deleted in the end of the program I get a segmentation fault in the destructor of the DetElement.
So please make sure, that there are no duplications and all DetElements are unique.

jhrdinka:
I was now testing the end caps again with your recent updates but I still do get the same volumeID (which is 0) for all sensitive component detector elements of the endcaps using this method (https://github.com/AIDASoft/DD4hep/blob/master/DDCore/include/DD4hep/DetElement.h#L397) after the geometry is fully built. Could you please check that?

vvolkl:
@jhrdinka I included a test in the geantino scan that resolves the Volume/Cell Ids and compares their positions with the ones from `PositionedHits`, so we are sure the volume manager works correctly for this geometry, for the simulation at least. This won't protect against duplicated volumes, but I'm confident that issue is resolved now as well. For now I disabled the ACTS ExtrapolationTest, because the DD4hep Plugin currently doesn't support the updated geometry, please re-enable this once the plugin is updated.

Note that there seems to be a bug in the DD4hep VolumeManager when the top level DetElement is set to type `compound`, it just doesn't find any volumes anymore. Maybe that is connected with the problems you saw earlier.

vvolkl:
Ah, I didn't see yet your comment, I will check that

jhrdinka:
I just wanted to add that the DetElement containing the Endcaps needs to be set to 'compound' (I added it myself locally to test). However if there is a bug in DD4hep this might be a problem - I could do a work around in ACTS. Did you create an issue in DD4hep, so that they can fix the bug?

vvolkl:
Ah there you go - as I said, the volume manager won't work anymore if you set it to `compound`. I didn't create an issue yet because I'm not sure it is a bug - why exactly is this needed? We don't set the disk DetElement to `compound` just because it has modules. Perhaps you can add an extension on this level as well?

jhrdinka:
As I understood 'compound' is meant to be set, if one has different detectors nested into one detecto(https://github.com/AIDASoft/DD4hep/blob/1e16d19e70bf45e6b3f096aa8c590c11fde5d633/DDCore/src/DetectorImp.h#L250). So if we have two end caps together it is fine to set it compound. For me I need to know which endcaps belong together. Since this was also done until now, because we used  the `DD4hep_SubdetectorAssembly` (which sets the parent DetElement to compound) to put together endcaps and barrel of one hierarchy, I do not understand why this did not happen until now. However as I said I can do a workaround with the extensions - then I will test if the problem with the volumeID which is not set still occurs and then we could open an issue in DD4hep.

Detector/DetFCChhTrackerTkLayout/src/TkLayoutEndcap_Geo.cpp
```diff
@@ -14,148 +13,169 @@ using DD4hep::Geometry::PlacedVolume;
...
   // definition of top volume
   std::string detName = xmlDet.nameStr();
   DetElement worldDetElement(detName, xmlDet.id());
+  DetElement posEcapDetElement(worldDetElement, "posEndcap", 0);
 
   Acts::ActsExtension::Config actsBarrelConfig;
-  actsBarrelConfig.isEndcap             = true;
+  actsBarrelConfig.isEndcap = true;
   Acts::ActsExtension* worldDetExt = new Acts::ActsExtension(actsBarrelConfig);
   worldDetElement.addExtension<Acts::IActsExtension>(worldDetExt);
```
jhrdinka: The Extension needs to be added to posEcapDetElement.

jhrdinka:
Could you please update the names of the subdetectors in the DD4hep sub detector assemblies (https://github.com/HEP-FCC/FCCSW/pull/252/files#diff-e0f9952f60edc5a6753f60a87d27c802R25) according to the changes you made?
For the forward endcap - you can either put it in a Subdetector assembly or I make an extra extension for the forward endcaps (- endcaps with no barrel on the same hierachy)

jhrdinka:
And the volumeIDs appear to be fine when the DetElement containing the Endcaps is not set `compound`. It is also fine when the EndCaps are put into a DD4hep Subdetector assembly (which also sets the mother volume to `compound`).
With updates in ACTS the detector builds fine now and tomorrow, after your update I will make final tests (no duplicated hits, all detector elements are found...).

Detector/DetComponents/src/CompareTrackHitPositionAndCellId.cpp
```diff
@@ -0,0 +1,132 @@
...
+  for (auto hit : (*hits)) {
+    long long int theCellId = hit.core().cellId;
+    debug() << theCellId << endmsg;
+    m_decoderBarrel->setValue(theCellId);
+    int system_id = (*m_decoderBarrel)["system"];
+    if (system_id == 0 || system_id == 1) {
+
+      auto detelement = volman.lookupDetElement(theCellId);
+      const auto& transformMatrix = detelement.worldTransformation();
+      double outGlobal[3];
```
jhrdinka: Please initialize the array

Detector/DetComponents/src/CompareTrackHitPositionAndCellId.cpp
```diff
@@ -0,0 +1,132 @@
...
+      debug() << "position: x: " << hit.position().x << "\t y: " << hit.position().y << "\t z: " << hit.position().z
+              << endmsg;
+      debug() << "position: x: " << edmPos.x << "\t y: " << edmPos.y << "\t z: " << edmPos.z << endmsg;
+      if ((abs(edmPos.x - hit.position().x) > l_tolerance) || (abs(edmPos.y - hit.position().y) > l_tolerance) ||
+          (abs(edmPos.z - hit.position().z) > l_tolerance)) {
+        error() << "positionedHit and position from CellId are not compatible!" << endmsg;
+        return StatusCode::FAILURE;
+      }
+
+    } else if (system_id == 2 || system_id == 3 || system_id == 29) {
```
jhrdinka: I am not such a fan of these hard coded system IDs. When I make the final readout update (just putting all readout information in one file, as discussed in one meeting), I could also introduce some global constants for that. 
Also `29` is not valid anymore.

Detector/DetComponents/src/CompareTrackHitPositionAndCellId.cpp
```diff
@@ -0,0 +1,132 @@
...
+      int layer_id = (*m_decoderBarrel)["layer"];
+      debug() << "\t layer " << layer_id;
+      int module_id = (*m_decoderBarrel)["module"];
+      debug() << "\t module " << module_id;
+      debug() << endmsg;
+
+      debug() << "Hit with Cell Id: " << theCellId << endmsg;
+      debug() << "position: x: " << hit.position().x << "\t y: " << hit.position().y << "\t z: " << hit.position().z
+              << endmsg;
+      debug() << "position: x: " << edmPos.x << "\t y: " << edmPos.y << "\t z: " << edmPos.z << endmsg;
```
jhrdinka: Could you please distinguish in the debug output which position is the hit positions and which is the edm position.

Detector/DetComponents/src/CompareTrackHitPositionAndCellId.cpp
```diff
@@ -0,0 +1,132 @@
...
+  double l_segGridSizeXEndcap = segmentationXZ->gridSizeX();
+  double l_segGridSizeZEndcap = segmentationXZ->gridSizeZ();
+  std::cout << l_segGridSizeXBarrel << std::endl;
+
+  DD4hep::Geometry::VolumeManager volman = m_geoSvc->lcdd()->volumeManager();
+
+  const fcc::PositionedTrackHitCollection* hits = m_positionedTrackHits.get();
+  double fcc_l1 = 0;
+  double fcc_l2 = 0;
+  double l_tolerance = 10;
```
jhrdinka: Why is the tolerance `10`? 
And could you please add units.

jhrdinka:
I like the new test now. This can also show us, in case the volume manager does not work. The only case which is missing is, as you said, duplicated volume entries, but I will double check that tomorrow.

vvolkl:
I wrote again to the DD4hep team - `compound` DetElements are used for Assemblies, which combine Detectors with different system ids. Therefore they don't have a VolID themselves. It's entirely expected that the volume manager won't work if we use them the way you suggested.

Detector/DetComponents/src/CompareTrackHitPositionAndCellId.cpp
vvolkl: Thanks for all the comments, it should be addressed in the commit below. I agree on the IDs, but since this is a fairly specific test I think it's ok in order to keep it simple. I still want to move the test to a more suitable directory ( we had some problem with the build system in `Test` ) but otherwise I think we can merge this.

jhrdinka:
Good. So I made tests now, with the update you made yesterday. 
I have just one tiny thing: please put the forward endcaps into an assembly - i thought adding an extra extension makes not too much sense, since all endcaps use the same detector consructor and you would just need it for the fwd endcaps, and anyway one can use the dd4hep subdetector assembly. However if you have another opinion on that, it is easy to add that.

With this little change all my tests succeeded:
- There are no duplicated hits in the geantino simulation anymore (i could probably add the root test script to fccsw) => there are no duplicated volumes anymore
- With the updated ACTS version the detector builds fine in ACTS and the extrapolation through the detector in ACTS shows me all sensitive detector elements correctly
-  In ACTS I also have this map where every detector element is mapped to its volumeID. Reading in the root file with the positioned hits of the geantino simulation I can find all detector elements from their cellID (after encoding to volumeID)

So from my side, after this little change is in, this can be merged.

vvolkl:
Hi Julia, If you have some tests, please add them, maybe in another PR.

Detector/DetComponents/src/CompareTrackHitPositionAndCellId.cpp
```diff
@@ -0,0 +1,139 @@
+#include "CompareTrackHitPositionAndCellId.h"
```
zaborowska: Shouldn't it be deleted from DetComponents? (it's under TestReco already?)

Detector/DetFCChhTrackerTkLayout/src/BeamTube_geo.cpp
```diff
@@ -12,6 +14,10 @@ static DD4hep::Ref_t create_element(DD4hep::Geometry::LCDD& lcdd, xml_h e, DD4he
   std::string det_name = x_det.nameStr();
   // Make DetElement
   DetElement beamtube(det_name, x_det.id());
+ Acts::ActsExtension::Config volConfig;
```
zaborowska: can you run checkformat?

Examples/options/geant_fullsim_pgun_geantino.py
```diff
@@ -35,6 +35,11 @@
                     outputs= ["SimG4SaveTrackerHits/saveTrackerHits" ],
                     eventProvider=pgun)
 
+from Configurables import CompareTrackHitPositionAndCellId
```
zaborowska: Maybe this can go to TestReco tests?

Reconstruction/RecTracker/CMakeLists.txt
```diff
@@ -37,6 +37,7 @@ gaudi_add_test(CombinatorialTrackSeeding
 			         FRAMEWORK options/combinatorialTrackSeedingTest.py
 			         DEPENDS Examples.GeantFullSimTrackerHits)
 
-gaudi_add_test(ExtrapolationTest
-               WORKING_DIRECTORY ${PROJECT_SOURCE_DIR}
-			         FRAMEWORK options/extrapolationTest.py)
+# TODO: to be re-enabled once ACTS DD4hepPlugin is up to date
```
zaborowska: maybe you can create an issue with this ToDo?

Test/TestReconstruction/src/CompareTrackHitPositionAndCellId.h
```diff
@@ -0,0 +1,45 @@
...
+// FCCSW
+#include "FWCore/DataHandle.h"
+class IGeoSvc;
+
+// datamodel
+namespace fcc {
+class TrackHitCollection;
+class PositionedTrackHitCollection;
+}
+
```
zaborowska: one line doxy comment that it's a test class linked to particular geometry?

Reconstruction/RecTracker/CMakeLists.txt
vvolkl: #260

vvolkl:
Thanks a lot for the comments, Anna! I tried to address the issues in the new commits, and the structure should indeed be a bit clearer.

PR #254, opened 2017-11-22
Save Geant TrackIDs with hits

vvolkl:

- Geant4 trackId saved with hit in `bits` field
- when merging pileup, an offset of `pileupVertexIndex * 2 **20` is added to the trackId to keep it unique

### DIFF
--- a/FWCore/src/components/PileupHitMergeTool.cpp
+++ b/FWCore/src/components/PileupHitMergeTool.cpp
@@ -82,11 +82,15 @@ StatusCode PileupHitMergeTool<Hits, PositionedHits>::mergeCollections() {
   Hits* collHitsMerged = new Hits();
   PositionedHits* collPosHitsMerged = new PositionedHits();
 
+  unsigned int collectionCounter = 0;
   for (auto hitColl : m_hitCollections) {
     // copy hits
     for (const auto elem : *hitColl) {
+      auto clon = elem.clone()
+      clon.bits += collectionCounter * 2 << 32; // add pileup vertex counter with an offset
       collHitsMerged->push_back(elem.clone());
     }
+    ++collectionCounter;
   }
   for (auto posHitColl : m_posHitCollections) {
     // copy positioned hits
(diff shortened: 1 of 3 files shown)


FWCore/src/components/PileupHitMergeTool.cpp
```diff
@@ -82,11 +82,15 @@ StatusCode PileupHitMergeTool<Hits, PositionedHits>::mergeCollections() {
   Hits* collHitsMerged = new Hits();
   PositionedHits* collPosHitsMerged = new PositionedHits();
 
+  unsigned int collectionCounter = 0;
   for (auto hitColl : m_hitCollections) {
     // copy hits
     for (const auto elem : *hitColl) {
-      collHitsMerged->push_back(elem.clone());
+      auto clon = elem.clone();
+      clon.bits(clon.bits() + collectionCounter * (2 << 20)); // add pileup vertex counter with an offset
```
jhrdinka: Here are my comments
- In this way you are limiting the number of particles to ~2M and the number of collection to ~2000 (from ATLAS numbers this is more than enough), however if in future this gets changed somehow, we have a problem. A possible consideration could be to separate the collection from the particle count
- Could you please not hard code the shift but specify it somewhere more visible and together with that document what you are doing, why you are doing it and what limitations we have because of that

vvolkl: Hmmm, I disagree, this does not limit anything at first - if you tried to simulate an event with full pileup at once instead of overlaying them you will have the same limitation. But I agree that if that happens, the trackID I'm saving to bits is not unique anymore, which could have hard to trace effects on efficiency calculations among other things. I'll add a check here that this doesn't happen, and let the algorithm fail hard in case it does. I don't think the offset needs to be moved anywhere, since it should be transparent to the user, and in case it does make problems, here is exactly the place where stuff needs to be changed.

Of course there are many ways this could be improved - adding a safe and convenient way to check what pileup event a hit belongs to for example. But the idea was to come up with a quick way so this information is there in the next release and any production campaigns we want to start soon after that. I would also change the edm at one point to make the field more descriptive, but I  propose to leave it for another PR. I will add some comments, though.

jhrdinka: Good, however you can move the offset e.g. in the header file and document it cleanly there. I think a hard coded shift somewhere in the source file can lead to inconsistencies and this would also be more clear for the user, than having it in a source file.

PR #255, opened 2017-11-24
Update simple tracker sd

jhrdinka:
Minor changes to SimpleTrackerSD
- at the moment the difference of the post step to the pre step point is set for the momentum. Now the momentum of the pre step point is used.
- Add a check if the track is still alive

### DIFF
--- a/Detector/DetSensitive/src/SimpleTrackerSD.cpp
+++ b/Detector/DetSensitive/src/SimpleTrackerSD.cpp
@@ -38,8 +38,6 @@ bool SimpleTrackerSD::ProcessHits(G4Step* aStep, G4TouchableHistory*) {
   if (edep == 0.) return false;
   // get track
   const G4Track* track = aStep->GetTrack();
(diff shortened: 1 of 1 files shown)


jhrdinka:
I added the comment now. About the momenum - when debugging I saw that it was not set correctly (PostStepPoint-PreStepPoint , for the direction of the momentum we would have needed the unit vector of that) and it is also correct, that I need it for digitization.

vvolkl:
Hi Julia,

Thanks for the updates, but the comment on the track status is not really what I meant - in my understanding of Geant4 this check isn't necessary and I wanted to know why it is needed. In fact I asked the Geant4 team and apparently we shouldn't do this. Please remove the `if` statement and check if you actually see dead tracks at this point in the program. If you do, please open an issue so it can be investigated properly.

jhrdinka:
Ok, I just got the tip from a colleague that we should check that. However when it is not needed at this point, then it should not be there - I removed it now.

PR #262, opened 2017-12-20
Set DD4hep print level from Gaudi msg level

zaborowska:

- Set the print level of DD4hep to output level of Gaudi (GeoSvc). This allows getting rid of most of the DD4hep output if GeoSvc.OutputLevel = WARNING


vvolkl:
@zaborowska I guess this requires a newer DD4hep version than we have in LCG 88?

zaborowska:
@vvolkl - no, just coffee ;)
In the new version, the change is only in the namespace DD4hep -> dd4hep.

PR #263, opened 2017-12-20
Modification of first layer of ecal barrel

zaborowska:

- start lead absorber from the second layer of inclined ECal barrel. This way energy deposited in the first layer which is an input to the correction for the upstream material is more uniform in the azimuthal angle.


zaborowska:
retest this please

vvolkl:
retest this please

PR #265, opened 2017-12-20
Switch Tracker beampipe for the centrally defined one

vvolkl:
The simulation plans require the full beampipe, instead of the one that was simplified for compatibility with ACTS. This should be harmonised with the update of the DD4hepPlugin


zaborowska:
Hi @vvolkl, is this PR actually still needed? Is the beampipe in ACTS ok now?

vvolkl:
It is, but it requires changes from @drasal in tklayout.

zaborowska:
Ok, thanks, I hope it will be addressed soon. I am closing this PR then.

PR #277, opened 2018-01-18
Calo topo clustering

CoralieNeubueser:
Needs PR# 245.

- new RecInterface that allows to get noise per Cell from cellID
- enables to search for larger neighbours range in det:utils:neighbours by 5th argument (int aSteps)

- CombinedCaloTopoCluster Algorithm that merges Cells into cluster
       - based on the ATLAS top clustering; 3 thresholds are set for the seed cell, the neighbouring cells and if you fail to find another neighbour in the last iteration, accept all found cells with energy > 0
       - arguments needed are 2 cell collections, 2 readouts, if noise was added also parse the noise tool, the 2 cell position tools for specific geometry, a neighbours range (if not set, set to 1), and the ID of the last ECal layer
       - the cluster in the ECal and HCal are merged in case of hit cells in the last ECal and first HCal layer, the matching is based on the cell positions in eta and phi and the segmentation the the HCAL in these dimensions
       - the algorithm needs the CellPositionsTools proposed in pr#245

To be done:

- Output collection for Cells that are not clustered


zaborowska:
ok to test

CoralieNeubueser:
The PR will be re-opened with an updated Clustering Algorithm.

PR #281, opened 2018-02-19
BField Service

jhrdinka:
This PR implements an fcc magnetic field service.

-`IBFieldService` interface with two implementations:
  1. `ConstantBFieldSvc` for a constant magnetic field (wrapping Acts implementation)
  2. `InterpolatedBFieldSvc` for a non constant magnetic field, with possibility to read in the magnetic field from a map in root or txt/csv  (wrapping Acts implementation).
- Example using the `BFieldWriter` to write out the magnetic field + a corresponding root script to display the field
- A Geant4 wrapper (`BFieldG4`, `SimG4BFieldTool`) which internally uses the `IBFieldSvc` to allow usage of the Magnetic field service during simulation 
- Tests of the magnetic field service for FCChh bField map:
  1. Consistency test if the magnetic field service returns the right values at the positions given by the field map 
  2. Symmetry test, since the FCChh magnetic field has z- and phi-symmetry
  3. Test if the geant4 wrapper gives the same values as the plain implementation


vvolkl:
Hi Julia,

I have more comments, but two general points first:

Having a math "library" for three `typedef`s bloats the framework and is confusing for new contributors (plus, shouldn't it use the eigen types directly instead of the ACTS wrappers?). I'd suggest you use the types you need directly, and we check if we need to refactor math code in a future PR.

I'm surprised at all the modules that now depend on ACTS. Is it possible to write a BFieldService that hides its implementation details (ACTS) to remove those dependencies?

jhrdinka:
Hi Valentin,

I once asked about implementing a Math library in FCCSW in a meeting and since there were no objections, I implemented it this way. I guess it will probably grow over time, however if wished I can remove it.

I know I also do not like all these dependencies. However I did not find a way in Gaudi-cmake to inherit the ACTS dependency for all submodules (like using the PUBLIC keyword in normal cmake). Anyway we have the same problem for all other external software packages we are using (e.g. ROOT, Geant4) and since it is our tracking package and we will also use it at other places, I think it is fine.

zaborowska:
@jhrdinka please have a look at failing tests as reported on [cdash](http://cdash.cern.ch/buildSummary.php?buildid=470780)

JavierCVilla:
@phsft-bot build please

JavierCVilla:
@phsft-bot build

zaborowska:
Hi @jhrdinka,
your root file is now in `/eos/project/f/fccsw-web/testsamples/FCChhBField_rz.root` together with all other samples, please change the path in all examples.

zaborowska:
I haven't noticed your txt file. You shouldn't exclude it from testing. It's copied into the same directory now.

jhrdinka:
Ok, I will enable the test again.

jhrdinka:
The reason why the number of calls between the Geant4 and Acts implementation had such a large difference was found and the Geant4-wrapper of the magnetic field service updated: now boundaries can be set.

Using the ConstBFieldSvc is ~7% slower than the pure const Geant4 implementation for 10000 single muon events. Using the FCChh magnetic field map with the InterpolatedBFieldSvc makes the simulation around a third slower. The simulation output can be found in the annex.

[ActsInterpolatedBField.txt](https://github.com/HEP-FCC/FCCSW/files/2804759/ActsInterpolatedBField.txt)
[ActsConstBField.txt](https://github.com/HEP-FCC/FCCSW/files/2804760/ActsConstBField.txt)
[G4ConstBField.txt](https://github.com/HEP-FCC/FCCSW/files/2804761/G4ConstBField.txt)

vvolkl:
Hi Julia,
thanks. Seeing this I realize we should actually use the `G4ExactHelixStepper` for the case of the constant B-field.

As discussed above, the interface for the magnetic field shouldn't use Acts types. Will revisit this after v1.0

vvolkl:
I don't think this is active anymore. I'll open an issue to keep track.

PR #285, opened 2018-02-25
Update SaveTrackerHits with Info necessary for Digitization

vvolkl:
This PR saves the Geant4 postStepPosition to edm and gets rid of the deprecated `dd4hep::sim::Geant4Hit` class in the process, introducing `fcc::Geant4PreDigiTrackHit` instead.

I also removed few tests in DetSensitive which were using dd4hep's `Geant4Tracker` SD, which I don't think we use - we might have slightly worse test coverage because of it, I'm open to suggestions for new tests!


JavierCVilla:
@phsft-bot build please

PR #286, opened 2018-02-26
Geometric Tracker Digitization

jhrdinka:
This merge request is a first implementation of geometric traker digitization for FCChh. 

-  The `Acts::DigitizationModules` later needed for digitization are created from the full detector description and handed over to Acts
- A helper struct `FCCDigitizationCell` which is used by the digitizer
- The `GeometricTrackerDigitizer` which creates a `fcc::TrackClustersCollection` from input `fcc::DigiTrackHitAssociationCollection` and writes it to the edm

Update:
- The digitization/clusterization was now optimized it terms of speed and memory consupmtion
- The clusterization (grouping cells together to clusters) was much too slow and was now replaced with a faster algorithm using hash maps. Two different implementation + their variations have been tested for that. The new implementation is unit tested here (https://gitlab.cern.ch/acts/acts-core/merge_requests/435).  In the next release it can be used directly from ACTS and the functions can be removed here. The new algorithm scales linearly with PU.
- Implemented hits merging similar as done in ATLAS and Belle (by Zbynek).

There are still some memory issues:
- Just reading in the hits (~18Mio) from a 1000PU event already takes 6GB of memory - when calculating how much memory these hits need from their data structure it should be less than 1GB 
=> Investigation ungoing for that from Benedikt
=> One part-solution is the hits merging implemented in this merge request. It merges the steps of Geant4 within the same module of the same particle to one hit only. This reduces the total number of hits, thus making memory consumption smaller and this will also fasten up the digitization, because there are less cells to be calculated.
- For future long term plan we should think of the possibility to read in hits by module - can be parallized and would be less data at once
- Creating the output edm track clusters takes another 2GB, when also creating their corresponding track hits (as done in this merge request) another 3GB are needed


zaborowska:
@jhrdinka please have a look at failing test on [cdash](http://cdash.cern.ch/viewTest.php?onlyfailed&buildid=471933)

jhrdinka:
This needs to be rebased anyway. Since we discussed yesterday, that this does not need to be in for the production yet, I marked it WIP.

zaborowska:
Hi Julia,

I noticed two issues which were already pointed out by Valentin in PR #281.

First one being the `typdef`s.  Please get rid of them from here as well.

The second aspect is including and linking against Acts all the submodules. What Valentin meant here is not that this should be done globally, but rather that packages that do not need Acts should have nothing in common with it. An example: `RecCalorimetry`.

For same reason, please move `Detector/DetCommon/DetCommon/DigiModuleHelper.h` etc. from `DetCommon` maybe to `Detector/Acts` or similar directory.

Thanks,
Anna

jhrdinka:
Hi Anna,

thanks for your comment, I will apply that.

Cheers,
Julia

PR #289, opened 2018-03-02
Cast mag_driver for compatibility with Geant4 10.4

vvolkl:
This implements a workaround for a problem with `RenewStepperAndAdjust` in Geant4 10.4


JavierCVilla:
@vvolkl I just tried it locally against Geant4-10.04 and it also requires to include a new header in order to use `G4MatInt_Driver`:

`#include G4MagIntegratorDriver.hh`

JavierCVilla:
Just to bear in mind: Merging this would require to use the `LCG_93` release and for that IMHO it would be better to be compatible with `DD4hep-01.05`

vvolkl:
Hi Javier, I think this solution is compatible works fine with the current stack, at least the tests are passing :)

JavierCVilla:
@vvolkl Completely right, I though the `init.sh` script had been modified.

PR #300, opened 2018-03-14
Specify user limits

zaborowska:

- implements regions with specified step limits (max step) or track limits (max track length, time etc)
- production cuts are not added, set to defaults of physics list

Validation has been done for a simple envelope (example) and by Valentin for tkLayout tracker.

### DIFF
--- a/Examples/options/geant_userLimits.py
+++ b/Examples/options/geant_userLimits.py
@@ -29,7 +29,7 @@ from Configurables import SimG4Svc, SimG4UserLimitPhysicsList, SimG4UserLimitReg
 from GaudiKernel.SystemOfUnits import mm
 ## create region and a parametrisation model, pass smearing tool
 regiontool = SimG4UserLimitRegion("limits", volumeNames=["TrackerEnvelopeBarrel"],
-                                       maxStep = 1)
+                                       maxStep = 1*mm)
 # create overlay on top of FTFP_BERT physics list, attaching fast sim/parametrization process
 physicslisttool = SimG4UserLimitPhysicsList("Physics", fullphysics="SimG4FtfpBert")
 # attach those tools to the G4 service
--- a/Examples/options/geant_userLimits.py
+++ b/Examples/options/geant_userLimits.py
@@ -1,0 +1,60 @@
+from Gaudi.Configuration import *
+
+# Data service
+from Configurables import FCCDataSvc
+podioevent = FCCDataSvc("EventDataSvc")
+
+from Configurables import HepMCFileReader, GenAlg
+readertool = HepMCFileReader("ReaderTool", Filename="/eos/project/f/fccsw-web/testsamples/FCC_minbias_100TeV.dat")
+reader = GenAlg("Reader", SignalProvider=readertool)
+reader.hepmc.Path = "hepmc"
+
+# reads an HepMC::GenEvent from the data service and writes a collection of EDM Particles
+from Configurables import HepMCToEDMConverter
+hepmc_converter = HepMCToEDMConverter("Converter")
+hepmc_converter.hepmc.Path="hepmc"
+hepmc_converter.genparticles.Path="allGenParticles"
+hepmc_converter.genvertices.Path="allGenVertices"
+
+# DD4hep geometry service
+# Parses the given xml file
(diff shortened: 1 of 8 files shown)


Examples/options/geant_userLimits.py
```diff
@@ -0,0 +1,60 @@
...
+                                         'file:Detector/DetFCChhBaseline1/compact/FCChh_TrackerAir.xml'],
+                                         OutputLevel=INFO)
+
+# Geant4 service
+# Configures the Geant simulation: geometry, physics list and user actions
+from Configurables import SimG4Svc, SimG4UserLimitPhysicsList, SimG4UserLimitRegion
+from GaudiKernel.SystemOfUnits import mm
+## create region and a parametrisation model, pass smearing tool
+regiontool = SimG4UserLimitRegion("limits", volumeNames=["TrackerEnvelopeBarrel"],
+                                       maxStep = 1)
```
vvolkl: Could you add unit ? I think you planned to, judging from import two lines above.

Sim/SimG4Full/src/components/SimG4UserLimitRegion.h
```diff
@@ -0,0 +1,63 @@
...
+private:
+  /// Regions used to set user limits
+  /// deleted by the G4RegionStore
+  std::vector<G4Region*> m_g4regions;
+  /// User limits
+  std::vector<std::unique_ptr<G4UserLimits>> m_userLimits;
+  /// Names of the volumes where user limits should be attached (set by job options)
+  Gaudi::Property<std::vector<std::string>> m_volumeNames{this, "volumeNames", {}, "Names of the volumes"};
+  /// max allowed Step size in this volume  (set by job options)
+  Gaudi::Property<double> m_maxStep{this, "maxStep", DBL_MAX, "maximum step"};
```
vvolkl: I know it's named like this in Geant, but with this naming, it's easy to confuse this variable for the maximum number of steps in the volume. So it's good you added the doxygen comment, I think with this it's ok! :+1:

Sim/SimG4Full/src/components/SimG4UserLimitRegion.cpp
```diff
@@ -0,0 +1,52 @@
...
+  G4LogicalVolume* world =
+      (*G4TransportationManager::GetTransportationManager()->GetWorldsIterator())->GetLogicalVolume();
+  for (const auto& trackerName : m_volumeNames) {
+    for (int iter_region = 0; iter_region < world->GetNoDaughters(); ++iter_region) {
+      if (world->GetDaughter(iter_region)->GetName().find(trackerName) != std::string::npos) {
+        /// all G4Region objects are deleted by the G4RegionStore
+        m_g4regions.emplace_back(
+            new G4Region(world->GetDaughter(iter_region)->GetLogicalVolume()->GetName() + "_userLimits"));
+        m_g4regions.back()->AddRootLogicalVolume(world->GetDaughter(iter_region)->GetLogicalVolume());
+        m_userLimits.emplace_back(new G4UserLimits(m_maxStep, m_maxTrack, m_maxTrack, m_minKineticEnergy, m_minRange));
```
vvolkl: We don't need a conversion from Gaudi to G4 units here?

vvolkl:
I checked this with the inner barrel of the  tkLayout tracker and it seems to be working nicely, I add the plots for completeness:
out_limits_digivalidation__steplengthhist_nocut
out_limits_digivalidation__steplengthhist_wcut

Sim/SimG4Full/src/components/SimG4UserLimitRegion.h
zaborowska: I can rename it, but I believe it will introduce even more confusion, I prefer to stick to Geant's nomenclature. For number we use `num` in the name (at least usually...).

zaborowska:
Thanks Valentin,
that's true, if we say we use Gaudi units from job options we should stick to it (concidentally they are the same as Geant's but I added the conversions anyway).
For completeness of your completeness I attach the plots from the example. `TrackerEnvelopeBarrel` has the region with step limit attached, while `FwdTrackerPos` has not.

steplength_1mmcut

steplength_nocut

PR #301, opened 2018-03-16
[WIP] Adding cluster and hit collection merge tool

CoralieNeubueser:

- PileupHitMergeTool accepting also only one HitCollection
- PileupClusterMergeTool to merge ClusterCollections

To be done:

- debug tests


Examples/options/overlaySignalAndPileup.py
```diff
@@ -45,13 +45,13 @@
 trackhitsmergetool = PileupTrackHitMergeTool("MyTrackHitMergeTool")
 # branchnames for the pileup
 trackhitsmergetool.pileupHitsBranch = "hits"
-trackhitsmergetool.pileupPositionedHitsBranch = "positionedHits"
+#trackhitsmergetool.pileupPositionedHitsBranch = "positionedHits"
```
vvolkl: This can stay, I think.

FWCore/src/components/PileupClusterMergeTool.cpp
```diff
@@ -0,0 +1,94 @@
...
+  return StatusCode::SUCCESS;
+}
+
+template <class Clusters>
+StatusCode PileupClusterMergeTool<Clusters>::readPileupCollection(podio::EventStore& store) {
+  // local pointers, to be filled by the event store
+  const Clusters* hitCollection;
+
+  // get collection address and store it in container
+  bool hitCollectionPresent = store.get(m_pileupClustersBranchName, hitCollection);
```
vvolkl: The variable names are out of sync with the code from copying, I think

FWCore/src/components/PileupClusterMergeTool.cpp
```diff
@@ -0,0 +1,94 @@
...
+  return StatusCode::SUCCESS;
+}
+
+template <class Clusters>
+StatusCode PileupClusterMergeTool<Clusters>::mergeCollections() {
+
+  // ownership given to data service at end of execute
+  Clusters* collClustersMerged = new Clusters();
+
+  unsigned int collectionCounter = 0;
```
vvolkl: If we want to keep track of the associations of hits to clusters, this will get a bit more complex, see
https://github.com/HEP-FCC/FCCSW/blob/master/FWCore/src/components/PileupParticlesMergeTool.cpp#L76

FWCore/src/components/PileupHitMergeTool.cpp
```diff
@@ -51,26 +51,29 @@ StatusCode PileupHitMergeTool<Hits, PositionedHits>::readPileupCollection(podio:
   }
 
   /// as above, for the positioned collection
-  bool posHitCollectionPresent = store.get(m_pileupPosHitsBranchName, posHitCollection);
-  if (posHitCollectionPresent) {
-    m_posHitCollections.push_back(posHitCollection);
-  } else {
+  if (!store.get(m_pileupPosHitsBranchName, posHitCollection)) {
```
vvolkl: I don't really know what's the best way to do it, but maybe it's good to be more explicit about the user intention? If I made an error with the collectionName I maybe would want the program to even exit with failure at this point. But if I don't want to add a signal collection then I don't need to be warned, no?

FWCore/src/components/PileupHitMergeTool.h
```diff
@@ -66,6 +66,9 @@ class PileupHitMergeTool : public GaudiTool, virtual public IEDMMergeTool {
 
   /// offset with which the pileup event number is added to the trackID 
   const unsigned int m_trackIDCollectionOffset = 2 << 20;
+
+  /// merge HitCollection only
```
vvolkl: Comment not very clear

PR #302, opened 2018-03-20
Update version of the FCC externals

JavierCVilla:
Setup the environment using the new version of external packages.


JavierCVilla:
@vvolkl I think both PR's are independent, this one updates the `init.sh` to run the CI builds against our latest release and hence latest externals. Otherwise we have a FCCSW version released and installed in `cvmfs` which was built against a stack different than the current CI builds. 

#303 requires a completely new release of the fcc externls built against `LCG_93`.

vvolkl:
Ok, I see, thanks!

PR #305, opened 2018-03-28
Only add trackid offset for tracker hits

vvolkl:
Fix for using the overlay algorithm on full detector events


jhrdinka:
Hello Valentin, thank you for this pull request.
Unfortunately the problem seems not to be solved with just shifting the track IDs. (However it is fine, since the IDs are not used in the calo).
The problem is, that the track IDs of only the tracker can go up to nearly 60 Mio for one event already. Therefore we will have duplicated IDs and it is a "silent" problem, which we will not notice.
screen shot 2018-03-22 at 11 45 06

jhrdinka:
Hi Valentin,
I am sorry, I did not see the hit counter. With that I guess it works. Anyway it is good to have it documented what the initial problem was above. I can test, if it is working now.

zaborowska:
Hi @jhrdinka, did you test if it works (as you mention in the comment above)?

vvolkl:
Closing this since MCTruth for a full 1000 pileup event is an edge case that is better handled differently. We should have an edm discussion at one point though

PR #312, opened 2018-05-04
Sliding window cluster corrections, estimation of pileup

zaborowska:

- cells are attached to the sliding window cluster reconstruction. However, the weight/fraction is not added, so in case the clusters share some cells, those cells are attached to both clusters.
- cluster correction algorithm: for eta position correction, as well as addition of corrections to the energy: for the upstream material correction and addition of the pileup noise (per cluster). Those were up to now addressed in the analysis/plotting macros.
- pile-up estimation in calorimetry. Algorithm that @faltovaj wrote to estimate cell noise levels got extended to the estimation of pileup noise per cluster. Several sizes of clusters can be specified.


Detector/DetFCChhCalDiscs/compact/Endcaps_coneCryo.xml
```diff
@@ -60,19 +60,19 @@
   </define>
   <readouts>
     <readout name="EMECPhiEta">
-      <segmentation type="FCCSWGridPhiEta" phi_bins="704" offset_phi="-pi+(pi/704.)" grid_size_eta="0.01" offset_eta="-2.515"/>
+      <segmentation type="FCCSWGridPhiEta" phi_bins="704" offset_phi="-pi+(pi/704.)" grid_size_eta="0.01" offset_eta="-2.51024"/>
```
CoralieNeubueser: Why did these numbers change?

Detector/DetFCChhCalDiscs/compact/Forward_coneCryo.xml
```diff
@@ -57,19 +57,19 @@
   </define>
   <readouts>
     <readout name="EMFwdPhiEta">
-      <segmentation type="FCCSWGridPhiEta" phi_bins="256" offset_phi="-pi+(pi/256.)" grid_size_eta="0.025" offset_eta="-5.9975"/>
+      <segmentation type="FCCSWGridPhiEta" phi_bins="256" offset_phi="-pi+(pi/256.)" grid_size_eta="0.025" offset_eta="-6.00024"/>
```
CoralieNeubueser: see question above.

Reconstruction/RecCalorimeter/src/components/CaloTowerTool.cpp
```diff
@@ -38,62 +38,136 @@ StatusCode CaloTowerTool::initialize() {
 
   // check if readouts exist & retrieve PhiEta segmentations
   // if readout does not exist, reconstruction without this calorimeter part will be performed
+  std::pair<dd4hep::DDSegmentation::Segmentation*, SegmentationType> tmpPair;
   info() << "Retrieving Ecal barrel segmentation" << endmsg;
-  m_ecalBarrelSegmentation = retrieveSegmentation(m_ecalBarrelReadoutName);
+  tmpPair = retrieveSegmentation(m_ecalBarrelReadoutName);
```
CoralieNeubueser: Could you add to the comments in CaloTowerTool.h, that either a Phi-Eta or Multi segmentation is required?

Sim/SimG4Components/src/SimG4SaveTrackerHits.cpp
```diff
@@ -67,7 +67,7 @@ StatusCode SimG4SaveTrackerHits::saveOutput(const G4Event& aEvent) {
       collect = collections->GetHC(iter_coll);
       if (std::find(m_readoutNames.begin(), m_readoutNames.end(), collect->GetName()) != m_readoutNames.end()) {
         size_t n_hit = collect->GetSize();
-        info() << "\t" << n_hit << " hits are stored in a tracker collection #" << iter_coll << ": "
+        verbose() << "\t" << n_hit << " hits are stored in a tracker collection #" << iter_coll << ": "
```
CoralieNeubueser: thanks for this!

Detector/DetFCChhCalDiscs/compact/Endcaps_coneCryo.xml
zaborowska: It changed to be consistent with the [offset of barrel](https://github.com/HEP-FCC/FCCSW/blob/master/Detector/DetFCChhECalInclined/compact/FCChh_ECalBarrel_Common.xml#L56). Without this change it matters if we run reconstruction of barrel only, or with endcaps + forward (there is a shift once endcaps&forward are included).

Detector/DetFCChhCalDiscs/compact/Forward_coneCryo.xml
zaborowska: again - consistency with ecal barrel. Otherwise we need to be careful to always run sim and reco on the same set of calorimeters.

Reconstruction/RecCalorimeter/src/components/CaloTowerTool.cpp
zaborowska: sure, but in fact it is always PhiEta that is required. Just sometimes it can be 'hidden' behind the multi segmentation (but each sub-segmentation should still have phi-eta, even though it may have different cell size)

Reconstruction/RecCalorimeter/CMakeLists.txt
```diff
@@ -72,6 +72,15 @@ gaudi_add_test(reconstructionFullCaloSystemWithNoise
                 FRAMEWORK tests/options/runFullCaloSystem_ReconstructionSW_noiseFromFile.py
                 PASSREGEX "Segmentation cells  \\(Nphi, Neta, minEta\\): \\[704, 335, 1\\]")
 
+gaudi_add_test(correctionFullCaloSystem
+                WORKING_DIRECTORY ${PROJECT_SOURCE_DIR}
+                DEPENDS reconstructionFullCaloSystemWithNoise
```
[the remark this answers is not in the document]
zaborowska: indeed, it's a mismatch. I updated the input file.

zaborowska:
Hi, 

thanks for checking this PR. I will rename the files according to Coralie's comment.

Regarding the cells & clusters, in the sliding window the approach would be to 'share' cells between clusters if they belong to more than 1 cluster (they are on a boundary of clusters). However, for the moment we have no way of bookkeeping the "shared" fraction of cell's energy within given cluster (that is normally calculated as ratio of area contained in the cluster) in podio. Therefore, in order to have the possibility of attaching cells to a cluster the "sharing" has to be switched off.

Other comments are addressed directly in the code.

Cheers,
Anna

vvolkl:
Thanks Anna!

PR #314, opened 2018-05-07
Fix Acts Extrapolation

vvolkl:
ACTS implementation of a Track Extrapolation Tool.
todo:
- [x] investigate small differences between acts and geantino extrapolation
- [x] switch extrapolation tool to more reasonable interface (return track states instead of acts extrapolationcell)
- [x] Add configurable constant BField


vvolkl:
canvas_1
canvas_1
@jhrdinka could you take a look if this discrepancy is connected to the tracking geometry?

vvolkl:
Added a Kalman Filter Prototype - The Extrapolation still needs to be fixed on the ACTS side

vvolkl:
Rebased on new externals. With the new acts version, the navigation is working now in the forward direction. The Kalman Filter Prototype I move to a new PR. This also includes a few updates to the geantino infrastructure, to allow for sensible test of the extrapolation.

PR #318, opened 2018-10-17
Commit hook

vvolkl:
Update of #217 

    Add a simple python script that calls git clang-format --diff:
        git clang-format --diff can be used to check the format of staged changes according to .clang-format
        If mis-matches are found it prints the diff and tells the user to run git clang-format
        git clang-format then actually does the re-format (see git clang-format -h after setting up the env for details). Opted for this less intrusive way to give the user a bit of control when things are actually changed
    Set up pre-commit hook in the init script by creating a symlink to the above script to FCCSW/.git/hooks/pre-commit


init.sh
```diff
@@ -2,6 +2,15 @@
 #!/bin/sh -u
 # set FCCSWBASEDIR to the directory containing this script
 export FCCSWBASEDIR="$( cd "$( dirname "${BASH_SOURCE[0]}" )" && pwd )"
+
+# setup the pre-commit hook that checks with clang-format before committing
+# slc6 git version is too old for use with clang-format
+export PATH=/cvmfs/sft.cern.ch/lcg/contrib/git/bin:$PATH
+export LD_LIBRARY_PATH=/cvmfs/sft.cern.ch/lcg/contrib/git/lib64:$LD_LIBRARY_PATH
+if [ ! -L $FCCSWBASEDIR/.git/hooks/pre-commit ]; then
+  ln -s $FCCSWBASEDIR/FWCore/scripts/pre-commit.py $FCCSWBASEDIR/.git/hooks/pre-commit
```
JavierCVilla:
```suggestion
   mkdir -p $FCCSWBASEDIR/.git/hooks
   ln -s $FCCSWBASEDIR/git-hooks/pre-commit.py $FCCSWBASEDIR/.git/hooks/pre-commit
```

JavierCVilla:
```suggestion
  mkdir -p $FCCSWBASEDIR/.git/hooks
  ln -s $FCCSWBASEDIR/git-hooks/pre-commit.py $FCCSWBASEDIR/.git/hooks/pre-commit
```

JavierCVilla: @vvolkl The code above should work.

Also, I would not make `git-hooks` a hidden folder.

vvolkl:
Thanks Javier! The top level directory is already a quite long list of packages and build configurations, in my opinion it distracts to have a folder there that in most cases won't be touched.

PR #319, opened 2018-11-01
Digitization1

jhrdinka:
This merge request the implementation of the geometric traker digitization for FCChh.

-The Acts::DigitizationModules later needed for digitization are created from the full detector description and handed over to Acts
-A helper struct FCCDigitizationCell which is used by the digitizer
-The GeometricTrackerDigitizer which creates a fcc::TrackClustersCollection from input fcc::DigiTrackHitAssociationCollection and writes it to the edm
Update:
-The digitization/clusterization was now optimized it terms of speed and memory consupmtion
The clusterization (grouping cells together to clusters) was much too slow and was now replaced with a faster algorithm using hash maps. Two different implementation + their variations have been tested for that. The new implementation is unit tested here ([https://gitlab.cern.ch/acts/acts-core/merge_requests/435]). In the next release it can be used directly from ACTS and the functions can be removed here. The new algorithm scales linearly with PU.
-Implemented hits merging similar as done in ATLAS and Belle (by Zbynek).
-Validation using Muons (see [https://indico.cern.ch/event/726333/contributions/3055261/attachments/1676170/2691172/ACTS_Status_26June18.pdf])
-Some writers to writing out cluster information which were and are used for studies
-Possibility to flag simulated particles which should be written out in the particle history: For some study I need the information of all particles which created a hit and using an energy cut for that selection was not suitable for me. The problem is that otherwise all particles are written out, even the ones which are created and die in non-sensitive material which are e.g. for just one muon a few hundred particles, which are not relevant for me. The flag can be set at runtime in the options for the tracker, so it does not affect anything, if not wished.

There are still some memory issues:

Just reading in the hits (~18Mio) from a 1000PU event already takes 6GB of memory - when calculating how much memory these hits need from their data structure it should be less than 1GB
=> Investigation ungoing for that from Benedikt
=> One part-solution is the hits merging implemented in this merge request. It merges the steps of Geant4 within the same module of the same particle to one hit only. This reduces the total number of hits, thus making memory consumption smaller and this will also fasten up the digitization, because there are less cells to be calculated.
For future long term plan we should think of the possibility to read in hits by module - can be parallized and would be less data at once
Creating the output edm track clusters takes another 2GB, when also creating their corresponding track hits (as done in this merge request) another 3GB are needed

Follow up: Despite the above problem the digitization runs without any problem for 1000PU and could be used for the studies for the CDR (see [https://indico.cern.ch/event/767683/contributions/3188745/attachments/1740269/2815858/OccupancyStudies.pdf]) . Still in future the memory consumption of the EDM should be checked.

Todo:
- Make own particle history action instead of changing current one (problem with pre and post tracking action)
- Use Segmentation directly from DD4hep using different segmentation for different layers (up to now, a work around was used)
- Update after update to new ACTS


vvolkl:
I don't think this is active anymore. I'll open an issue.

PR #320, opened 2018-11-08
Use the updated podio-ROOTReader to allow chaining input files

vvolkl:
Depends on https://github.com/HEP-FCC/podio/pull/66

Minor changes propagated from Podio, to allow using `TChain`ing input files. This PR changes the interface of the FCCDataSvc and breaks thus  many existing configuration files. Q for the reviewers, it is better to keep the `filename` option?


init.sh
```diff
@@ -13,4 +13,6 @@ if [ ! -L $FCCSWBASEDIR/.git/hooks/pre-commit ]; then
 fi
 
 weekday=`date +%a`
+
 source /cvmfs/fcc.cern.ch/sw/views/releases/externals/94.0.0/x86_64-slc6-gcc62-opt/setup.sh
+export CMAKE_PREFIX_PATH=/afs/cern.ch/work/v/vavolkl/public/fcc.cern.ch/sw/dev/podio:$CMAKE_PREFIX_PATH
```
JavierCVilla: This line should be removed.

FWCore/src/components/PileupOverlayAlg.cpp
```diff
@@ -26,7 +26,7 @@ StatusCode PileupOverlayAlg::initialize() {
   if (m_doShuffleInputFiles) {
     m_pileupFileIndex = static_cast<unsigned int>(m_flatDist() * m_pileupFilenames.size());
   }
-  m_reader.openFile(m_pileupFilenames[m_pileupFileIndex]);
+  m_reader.openFiles({m_pileupFilenames[m_pileupFileIndex]});
```
JavierCVilla: Do we want to use the old interface for single files or this one?

vvolkl: Here it doesn't really matter, no? actually the pileup infrastructure implements a simple version of TChain, so with this change it can be simplified in the future.

init.sh
vvolkl: yes. can you remind me when we have externals with the new podio available? Then I'll change it. Thanks!

JavierCVilla: Newest `podio` (current master) can be tested with today's nightlies:

```
/cvmfs/fcc.cern.ch/sw/views/nightlies/externals/latest/x86_64-slc6-gcc62-opt/setup.sh
```

We can modify the `init.sh` file to point to that `setup.sh` and if it works, we can consider a new release of the externals with a new `podio` version and other packages if needed.

FWCore/src/components/PileupOverlayAlg.cpp
JavierCVilla: Fair enough.

PR #322, opened 2018-12-17
Introduce Charginos and Neutralinos + their physics

jhrdinka:

- Chargino, Neutralino definitions with mass, width and lifetime not defined
- Geant4 Physicslist which defines the Charginos and Neutralinos and the decay of the Charginos into the Neutralion and a pion
- CharginoPhysicsTool which can be configured and hands over the relevant parameters of the charginos and neutralino to the physics list


vvolkl:
Hi Julia.
very nice thanks. This should be tested/validated/documented though. Can you at least point to study that uses this physics list? Also since they are not needed for most FCC simulations, please move these files to a separate subdirectory, for example `Sim/SimG4Susy`.

PR #324, opened 2019-02-19
fccrun: new entrypoint to FCCSW

vvolkl:
Adds a new command line utitlity `fccrun` that automatically parses the properties of gaudi configurables and creates argparse arguments from them. Hopefully self-documenting. Example:

```
~/repo/FCCSW (feature-fccrun)$ fccrun  ~/repo/FCCSimJobs/test/test.py -h
/home/vali/repo/FCCSimJobs/test/test.py
 -->  GenAlg  -->  HepMCToEDMConverter  -->  StableParticleFilter  -->  out  

usage: fccrun [-h] [--dry-run] [-v] [-n NUM_EVENTS] [-l] [--gdb]
              [--ncpus NCPUS] [--PdgCodes PDGCODES [PDGCODES ...]]
              [--EtaMin [ETAMIN]] [--EtaMax [ETAMAX]] [--PhiMax [PHIMAX]]
              [--PhiMin [PHIMIN]] [--PtMin [PTMIN]] [--PtMax [PTMAX]]
              [--logSpacedPt [LOGSPACEDPT]] [--accept ACCEPT [ACCEPT ...]]
              [--hepmcStatusList HEPMCSTATUSLIST [HEPMCSTATUSLIST ...]]
              [--input [INPUT]] [--tVertexMean [TVERTEXMEAN]]
              [--tVertexSigma [TVERTEXSIGMA]] [--zVertexMean [ZVERTEXMEAN]]
              [--yVertexSigma [YVERTEXSIGMA]] [--xVertexMean [XVERTEXMEAN]]
              [--yVertexMean [YVERTEXMEAN]] [--zVertexSigma [ZVERTEXSIGMA]]
              [--xVertexSigma [XVERTEXSIGMA]]
              [--outputCommands OUTPUTCOMMANDS [OUTPUTCOMMANDS ...]]
              [--filename [FILENAME]]
              [config_files [config_files ...]]

Run job in the FCC framework

positional arguments:
  config_files          Gaudi config (python) files describing the job

optional arguments:
  -h, --help            show this help message and exit
  --dry-run             Do not actually run the job, just parse the config
                        files
  -v, --verbose         Run job with verbose output
  -n NUM_EVENTS, --num-events NUM_EVENTS
                        Number of events to run
  -l, --list            Print all the configurable components available in the
                        framework and exit
  --gdb                 Attach gdb debugger
  --ncpus NCPUS         Start Gaudi in parallel mode using NCPUS processes. 0
[...]

PR #330, opened 2019-03-08
Release v0.10.0

vvolkl:
With the new TileCal implementation in, it seems like a good time to make a tag. I would propose to do v0.10 after the following PRs are in:

#324 
#320 
#312 
(optionally also the gaudi base configuration module I haven't managed to push yet)


vvolkl:
All of those are in - @JavierCVilla @clementhelsens I suggest we make the release!

PR #331, opened 2019-03-12
Geant4CaloHit fix and Jenkinsfile 

vvolkl:
The data type used in the  SimpleCaloSD sensitive detector was  planned to be removed in dd4hep, so I switch to a fcc type. This should be completely analogous to the tracker where it was done already a while ago.  

Also I save the trackID of CaloHits in the currently unused bits field, which is useful for the CaloHitMatching I am doing.

Includes my Jenkinsfile because I forgot to remove it and  should anyway be used by the CI very soon.


vvolkl:
@phsft-bot build

Detector/DetSensitive/src/AggregateCalorimeterSD.cpp
```diff
@@ -45,11 +45,11 @@ bool AggregateCalorimeterSD::ProcessHits(G4Step* aStep, G4TouchableHistory*) {
   dd4hep::Position pos(midPos.x(), midPos.y(), midPos.z());
```
JavierCVilla: Is `pos` still needed?

Detector/DetSensitive/src/AggregateCalorimeterSD.cpp
```diff
@@ -58,9 +58,14 @@ bool AggregateCalorimeterSD::ProcessHits(G4Step* aStep, G4TouchableHistory*) {
...
   // if not, create a new hit
   // deleted in ~G4Event
-  hitMatch = new dd4hep::sim::Geant4CalorimeterHit(pos);
+  hitMatch = new fcc::Geant4CaloHit(0, // track->GetTrackID()
+                                    0, // track->GetDefinition()->GetPDGEncoding()
+                                    edep,
+                                    0 // track ->GetGlobalTime()
+                                    ) ;
+
+  hit->position = midPos;
```
JavierCVilla: Is this `hit` related to `hitMatch` at this point?

vvolkl: good catch, this should be hitMatch

Detector/DetSensitive/src/AggregateCalorimeterSD.cpp
vvolkl: no

PR #345, opened 2019-05-22
Calorimeter Reconstruction Bugfixes

### LINKED ISSUE #342: Potential Integer overflow in Calo-Reconstruction Code
Here: <https://github.com/HEP-FCC/FCCSW/blob/master/Reconstruction/RecCalorimeter/src/components/LayeredCaloTowerTool.cpp#L219> the absolute value of the difference of two unsigned integers is calculated. Should be signed integers if used in this way.

### LINKED ISSUE #344: Bug: Use of CellIDs as CaloCell keys corrupted by RedoSegmentation
Quick documentation of some major debugging effort: In the calorimeter reconstruction energy deposits get accumulated in calorimeter cells. Internally the cells are implemented as a `std::unordered_map` before being written as a collection of EDM CaloHits. @CoralieNeubueser noticed that sometimes the number of  CaloCells would grow when adding hits even though the map should already complete (it is created by a dedicated geometry tool), like this:

```
...
CreateECalBarre...  DEBUG Number of calorimeter cells after merging of hits: 1765632
...
CreateECalBarre...  DEBUG Input Hit collection size: 1
CreateECalBarre...  DEBUG CellID : 439567255557is in map:0
CreateECalBarre...  DEBUG    system:5,cryo:0,type:0,subtype:0,layer:1,eta:264,phi:613
CreateECalBarre...  DEBUG Number of calorimeter cells after merging of hits: 1765633
CreateECalBarre...  DEBUG Output Cell collection size: 1765633
``` 
We found that in the map there was already a key that corresponds to the cell where the hit should be added.  All individual fields as seen in the decoded Valuestring are the same, nevertheless the full CellIDs are not equal.
This seemed impossible at first, but is in fact possible (although obviously not desired) because the readout in question does not use all 64 bits.
Thus any leading bits outside of the field definition can be changed without affecting how the cellID is converted to the fields defined in the readout.

How did the unused bits of the cellID get modified in the first place, though?
This does not happen with direct products of Geant4 simulation and their  `SaveOutput` tools.  
But the Calo reconstruction workflow includes a `RedoSegmentation` step that effectively rewrites the cellIDs. The CellId that are supposed to change from referring from one readout  to referring from another  were not completely cleared, but rewritten.
As old readout (`EcalBarrelEta`) is longer than the the new one (`EcalBarrelPhiEta`), leading ones from the eta field can remain in the cellID.

vvolkl:
Closes #344 and Closes #342 

See the discussion in the issue.

For `RedoSegmentation` it would be good to add a dedicated test, will follow in a future PR.

### DIFF
--- a/Detector/DetComponents/src/RedoSegmentation.cpp
+++ b/Detector/DetComponents/src/RedoSegmentation.cpp
@@ -94,7 +94,7 @@ StatusCode RedoSegmentation::execute() {
     // factor 10 to convert mm to cm
     dd4hep::DDSegmentation::Vector3D position(hit.position().x / 10, hit.position().y / 10, hit.position().z / 10);
     // first calculate proper segmentation fields
-    dd4hep::DDSegmentation::CellID newCellId = m_segmentation->cellID(position, position, volumeID(hit.cellId()));
+    dd4hep::DDSegmentation::CellID newCellId = m_segmentation->cellID(position, position, 0);
     // now rewrite all other fields (detector ID)
     for (const auto& detectorField : m_detectorIdentifiers) {
       oldid = m_oldDecoder->get(cellId, detectorField);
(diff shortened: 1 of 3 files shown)


clementhelsens:
Thanks Valentin, that looks good to me, I'll merge

PR #346, opened 2019-05-29
I/O of std::vectors and DriftChamber cleanup

vvolkl:
Patch of DataHandle so that it is finally possible to write elementary types in the framework. Major cleanup of Driftchamber reco code, uses the new functionality to get rid of an ad-hoc TFile used for the same purpose.


vvolkl:
@phsft-bot build

JavierCVilla:
@phsft-bot build

PR #351, opened 2019-07-01
Two particles in Calo Reco Tests

vvolkl:
Changes the test setup for calorimeter reconstruction so there is each time an electron in the forward region and a pion in the barrel region in order to have a more reliable test.


vvolkl:
Hi @clementhelsens,
no particular reason, it was mostly to be sure that both the forward and barrel collections are not empty, since otherwise the reconstruction for those is not tested.

I can add another test for that case!

vvolkl:
@phsft-bot  build

PR #353, opened 2019-07-11
Add CXX configuration in line with Gaudi config

JavierCVilla:
Allow users to define the cxx standard through the cmake configuration with: `-DCMAKE_CXX_STANDARD=<standard>` and propagate changes to Gaudi. 

This aligns the CMake configuration of FCCSW with the rest of cmake projects,

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -15,18 +15,41 @@ endif(SAS)
 find_package(GaudiProject)
 #---------------------------------------------------------------
 
+# Set up C++ Standard
+# ``-DCMAKE_CXX_STANDARD=<standard>`` when invoking CMake
+set(CMAKE_CXX_STANDARD 14 CACHE STRING "")
+
+if(NOT CMAKE_CXX_STANDARD MATCHES "14|17")
+  message(FATAL_ERROR "Unsupported C++ standard: ${CMAKE_CXX_STANDARD}")
+endif()
+
+# Prevent CMake falls back to the latest standard the compiler does support
+set(CMAKE_CXX_STANDARD_REQUIRED ON)
+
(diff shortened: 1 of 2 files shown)


JavierCVilla:
@phsft-bot build centos7/gcc62

JavierCVilla:
> why is the setup script called this_fccsw.sh and not setup.sh?

AFAIK `this_fccsw.sh` is supposed to be the equivalent to `this_root.sh` while `init.sh` or `setup.sh` are more related to the externals/environment.

gganis:
Well, thisroot.sh is very unconventional (in the spirit of ROOT ...), and all packages providing such a (very useful) script call it setup.sh . What it setups is defined by what comes before. I would expect/look for a setup.sh when setting up something, but I do not feel strong about this.

vvolkl:
@gganis indeed, I felt like setup implies more the environment: compiler versions, external packages etc, whereas `this_fccsw.sh` is somehow descriptive; the user might have some other installations of fccsw and chooses to use this one to execute his job.   In any case, no objections to calling it `setup.sh`, keeping to standard naming schemes is always good.

Continuing this line of thought: our lives may be even simpler if we drop the distinction of externals and framework, and install everything together. This was done for two reasons: for production campaigns we wanted to be able to have new installations in the framework quickly, i.e. in around one day.  But with the views I actually see no reasons why it should take longer to install everything: all the externals can be taken from an old installation  and will only be linked. Is that right Javier?
The other reason was that in the early days, FCCSW was not very useful on its own, but now I feel we enter a phase where for some applications at least, not much framework code needs to be changed (and users can write new components also as a depending gaudi project, like here: https://github.com/vvolkl/Geant4-exampleB4-inFCCSW)

PR #354, opened 2019-07-11
Add CMAKE_PREFIX_PATH as a hint in find modules

JavierCVilla:
Currently `FindFCCEDM` and `FindPODIO` rely on `FCCEDM_DIR` and `PODIO_DIR` variables to find `fcc-edm` and `podio` respectively, these variables might not be set though.

This PR adds the standard variable `CMAKE_PREFIX_PATH` as a hint to find the directories inside the `Find<package>` modules. 

Nonetheless this is a workaround, these `Find` modules may not be needed since in both cases CMake already generates the corresponding `fccedm-config.cmake` and `podio-config.cmake` files in the installation directories with the required information so, when properly configured, both `Find` modules will be redundant.

### DIFF
--- a/cmake/FindFCCEDM.cmake
+++ b/cmake/FindFCCEDM.cmake
@@ -1,4 +1,4 @@
-set(searchpath ${FCCEDM_DIR} ${PODIO_DIR}  )
+set(searchpath ${FCCEDM_DIR} ${PODIO_DIR} ${CMAKE_PREFIX_PATH} )
 
 # first find the FCC EDM
 find_library(FCCEDM_LIBRARY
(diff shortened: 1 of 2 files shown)


vvolkl:
I'm a bit surprised we even have `Find`-scripts for these packages. As you say, both fcc-edm and podio should have the right `config` files and properly export their targets  -- can you maybe try removing these two files? They should still be found.

JavierCVilla:
> can you maybe try removing these two files? They should still be found.

I did, the problem is that the corresponding `config` files are named: `fccedmConfig.cmake` and `podioConfig.cmake` (in lowercase) while we look for `FCCEDM` and `PODIO` packages in the FCCSW project (in uppercase), as a result that solution might require changes in the three packages.

I would prefer to do those changes on a different PR so that at least we are able to properly build FCCSW with Spack.

vvolkl:
Ok, we can definitely do that in another PR. But as far as I remember the Gaudi CMake configuration converts all package names to lowercase anyway.

PR #359, opened 2019-08-05
Build System: Install Example Option Files

vvolkl:
The job options have been distributed via the repository until now ("go to your clone of FCCSW and execute `fccrun Examples/options/...`"). Installing them hopefully makes it easier to write documentation.


gganis:
@vvolkl, does it install under 'share' or examples?

vvolkl:
Hi @gganis, at the moment still under "Examples". I did a quick check and to me the best thing seems to be to use `CMAKE_INSTALL_DATADIR` from `GNUInstallDirs` (which just installs everything under `share/Examples`). Updating now!

Detector/DetCommon/CMakeLists.txt
```diff
@@ -18,7 +18,7 @@ gaudi_add_library(DetCommon
                  LINK_LIBRARIES GaudiKernel DD4hep ROOT Geant4 DetSegmentation
                  PUBLIC_HEADERS DetCommon)
 
-install(DIRECTORY ${CMAKE_CURRENT_LIST_DIR}/compact DESTINATION Detector/DetCommon)
+install(DIRECTORY ${CMAKE_CURRENT_LIST_DIR}/compact DESTINATION ${CMAKE_INSTALL_DATADIR}/Detector/DetCommon)
```
JavierCVilla: The safest option to avoid possible clashes with other packages would be to use:

```
${CMAKE_INSTALL_DATADIR}/${CMAKE_PROJECT_NAME}/Detector/DetCommon
```

given that `CMAKE_INSTALL_DATADIR` is the same for all packages. Adding the project name we avoid issues with the folders of other projects using the same name.

vvolkl: Sure, that's a good suggestion.

PR #360, opened 2019-08-07
Prevent grep from existing in case of nomatch

JavierCVilla:
The current `init.sh` script fails if the internal `grep` command does not find any match. This avoids this situation so builds in SLC6 executed with `bash -e` do not fail.

### DIFF
--- a/init.sh
+++ b/init.sh
@@ -1,5 +1,13 @@
-
 #!/bin/sh -u
+
+# Set up the environment to build and test FCCSW
+# This script relies on CVMFS and the FCC Externals
+#
+# Usage:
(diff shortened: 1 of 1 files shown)


JavierCVilla:
@phsft-bot build

JavierCVilla:
These errors are most likely not related to the changes on this PR but to different issues in the jenkins nodes.

JavierCVilla:
@phsft-bot build

JavierCVilla:
@phsft-bot build just on centos7/gcc62

JavierCVilla:
@phsft-bot build

PR #372, opened 2019-10-17
Topo-cluster calibration/splitting and cone selection, used in simulation sample production /v04

CoralieNeubueser:

- cluster calibration: based on benchmark method, calibrating the HCal to hadronic scale and correcting for energy losses in cryostat between E&HCal
- cluster splitting based on ALTAS principle [1]
- cone selection of cells within cone around single particles, reduces reconstruction time significantly when running in pile-up conditions
- tools (cell positions/noise) adjusted for different TileCal description
- included new CellPositionsDummyTool, necessary with new Gaudi version.

[1] https://arxiv.org/abs/1603.02934


vvolkl:
Hi Coralie,

excellent, thanks! It seems your PR is missing the module `FCChhTkLayoutResolutionFormula` that you use in the `coneSelection` tests, can you include that?

PR #374, opened 2019-11-17
Remove test dependencies on HepMC text data files on EOS

vvolkl:
The file `/eos/project/f/fccsw-web/testsamples/FCC_minbias_100TeV.dat` which was used in a few tests seems to have been responsible for issues with the CI. This PR substitutes the reading particles from this file with a simple particle gun, which I think doesn't affect test coverage and should speed up things. Also, now it should be possible to run the tests without having to mount eos.


vvolkl:
@phsft-bot build please

PR #375, opened 2019-11-29
Fix/Improve JobOptionsSvc 

vvolkl:
The JobOptionsSvc currently saves all Properties of Components as configured in a job option file to the output root file, but it is missing the `ApplicationMgr`.
This PR adds this and some fixes to make the stored job options complete and executable.

See also here:  https://gitlab.cern.ch/gaudi/Gaudi/issues/95


vvolkl:
@phsft-bot  build also on centos7/gcc62

PR #376, opened 2019-12-03
Powheg and ResonanceDecayFilter Hooks for Pythia

vvolkl:
As needed for generation of some NLO processes, based on pythia examples and cmssw pythia8 interface.


selvaggi:
Seems ok to me. Thanks!

PR #377, opened 2019-12-11
[FWCore] Allow I/O of integers and floats

vvolkl:
Originally our implementation of the DataSvc only allowed for podio collections to be written to disk. This apparently sometimes didn't cover all the needs of developers, and people resorted to bypassing the DataSvc by having their algorithms write to their private `TFile`s.  I previously extended the DataSvc to be able to handle any objects for which there exists a ROOT dictionary. Here I make the final step and add primitive integers  and floating point types (which are handled a bit differently by root).

Strictly speaking it was already possible to write floats by using the deliberately vague type `fcc::FloatValue`. But in my opinion, when working with the produced datafiles, the simplicity of dealing with a branch of integers (for cuts in TTree::Draw etc) is really important, and even dealing with this simple struct wasn't quite the same.

There is some danger that floats and integers will be used for everything instead of the foreseen edm types, but I think we can simply call this out during code reviews when we see a missed opportunity to use a type that is better suited.

The code in DataHandle is becoming a bit more convoluted, but I tried to add comments to keep it clear.


vvolkl:
@phsft-bot build please

PR #385, opened 2020-04-02
Remove AFS Paths in Tests

vvolkl:
Moving some test samples to eos / the repository.


vvolkl:
@phsft-bot  build

vvolkl:
Hi @clementhelsens @gganis I think there was a change in the access settings for /eos/experiment/fcc, I cannot explain otherwise why the tests kept failing. I tried to be consistent with using xrootd for any files used in tests, so they can be used even if /eos is not mounted.

gganis:
Uhm ... /eos/experiment/fcc is open and I am not aware of any changes. But looking at the files I see changes removing /eos/project/f/fccsw-web. That requires at least a CERN token. On which machines was this failing? Can you sent me a pointer to the failing job(s) console output (on Jenkins)?

PR #390, opened 2020-06-01
Update to LCG release 97_FCC_2

vvolkl:

PR #391, opened 2020-06-02
Use K4FWCore in FCCSW

vvolkl:
Incorporates #390


vvolkl:
Part of the edm4hep transition branch now.

PR #393, opened 2020-06-28
Python3 Compatibility in fccrun

vvolkl:


clementhelsens:
@vvolkl shall we merge? or wait the CI is fixed?

vvolkl:
I think it's ok, the CI might be connected with the move to the new jenkins instance.

PR #395, opened 2020-09-05
First implementation of the FCC-ee CLD detector (from ILCSOFT, FCCee_o1_v04)

EmanuelPerez:

- Files that describe the CLD detector have been added into a new DetFCCeeCLD directory
- Files that describe the beam-pipe, the compensating and screening solenoids, the  HOM absorbers, the LumiCal, have been put into a new DetFCCeeCommon directory.

The starting point was the FCCee_o1_v04 model from ILCSoft. Minor changes were needed:
- use “SimpleTrackerSD” and “SimpleCalorimeterSD” as sensitive detector types (instead of "tracker" and "calorimeter")
- removed the part of the code that was using the lcio “encoders”
- in the XMLs describing the vertex detector and the tracker, the support structure has been commented (the corresponding driver code does not compile)

To be done:

- clean-up of DetFCCeeIDEA : for the elements that are in common, should use DetFCCeeCommon and not duplicate the files. I will submit a separate PR for that.


clementhelsens:
@phsft-bot build

Detector/DetFCCeeCLD/src/GenericCalBarrel_o1_v01_geo.cpp
```diff
@@ -0,0 +1,293 @@
...
+    
+    
+    //FOR NOW, USE A MORE "SIMPLE" VERSION OF EXTENSIONS, INCLUDING NECESSARY GEAR PARAMETERS
+    //Copied from Frank's SHcalSc04 Implementation
+    sdet.addExtension< LayeredCalorimeterData >( caloData ) ;
+    
+    return sdet;
+}
+
+DECLARE_DETELEMENT(GenericCalBarrel_o1_v01, GenericCalBarrel_o1_v01_geo )
```
andresailer, line 293: Please give them a different name to avoid clashes with the lcgeo library

andresailer: Actually, if only the function name of the creator is changed, please push the detectors to DD4hep/DDDectors instead.

clementhelsens: @andresailer I'm not sure I follow this comment. you suggest that @EmanuelPerez push this detector to DD4hep, but you will keep it in lcgeo?
I believe we start having too many places with DD4hep code, and we really need to have a discussion on how we want to deal with this for the long term.

andresailer: No, push it to dd4hep, and then we remove it from lcgeo eventually after marking it deprecated. Except that for DD4hep the detector name should be prefixed by `DD4hep_`

EmanuelPerez: Hi Andre ! Actually, the lcgeo code needs a minor change anyway, to use a sensitive detector type that is defined in Detector/DetSensitive (in the lcgeo code, the type "calorimeter" is hard-coded). I have changed the name of the GenericCalXX code to CLD_ GenericCalXX, and modified the code such that the sensitive type is taken from the xml, instead of being hard-coded. This requires changes in the xml. I could also push this code into DD4Hep if deemed useful (changing the name to DD4hep_GenericCalXX).

vvolkl (changes requested):
Hi @EmanuelPerez,

Thanks for this PR! One thing to consider is that in key4hep we discussed going the other direction, that is using more a central repository for detector geometries like lcgeo.

On the one hand it is nice to integrate this more in FCCSW, and I would certainly keep the folder DetFCCeeCLD and the the master xml file, on the other hand the continuity of things in lcgeo can also be valuable (even if the repository name isn't really accurate).

Using detector constructors and compact files from lcgeo in FCCSW is also very easy, as we are building everything in a common stack and it requires just prefixing `$LCGEO` to the paths for the xmls.

Nevertheless, I'm fine with merging this -- we can still deduplicate things in the future -- but I would suggest to keep the naming scheme oX_vY and maybe advance to v05. Let me know if you want to do it here, or I can do that in another PR.

PR #398, opened 2020-09-16
[cmake] fix setup script path order

vvolkl:
Prepend paths in the setup script instead of appending them, in order to pick up this install first.


gganis:
Looks good to me

PR #399, opened 2020-09-16
[delphes] save trkcov information

vvolkl:

PR #400, opened 2020-09-19
[delphes] save trkcov information (PR #399) + fix for read-only access compilation failure

gganis:
See PR #399  .

This adds a fix for the CI failure in preparation for the v0.14 tag


vvolkl:
Thanks @gganis - as discussed offline, we merge the #399

PR #403, opened 2020-09-25
FcceeLAr code cleaning

faltovaj:
- Updated dimensions for DetFCCeeIDEA-LAr
- Code cleaning in DetFCCeeECalInclined, files not needed removed
- Examples in RecFCCeeCalorimeter cleaned - removing parts of code not used


Detector/DetFCCeeECalInclined/compact/FCCee_ECalBarrel_calibration.xml
```diff
@@ -1,31 +1,117 @@
...
-      Passive plate inlcude lead in the middle, with steal on both sides, glued together.
-      Passive plates are inclined by a certain angle from the radial direction. The barrel is filled with liquid argon.
-      It includes cryostat.
+      Common settings for the inclined EM calorimeter.
+      The barrel is filled with liquid argon. Passive material includes lead in the middle and steal on the outside, glued together.
+      Passive plates are inclined by a certain angle from the radial direction.
+      In between of two passive plates there is a readout.
+      Space between the plane and readout is of trapezoidal shape and filled with liquid argon.
+      Definition of sizes, visualization settings, readout and longitudinal segmentation are specified.
       This file is used in the calculation of the sampling fraction that corrects energy for the energy depoisted in the absorber.
```
vvolkl, line 19:
```suggestion
      This file is used in the calculation of the sampling fraction that corrects energy for the energy depoisted in the absorber.
```
```suggestion
      This file is used in the calculation of the sampling fraction that corrects energy for the energy deposited in the absorber.
```

vvolkl (approved):
Thanks, Jana. Initially I wanted to keep the structure of hh for the compact files, but it makes sense to remove the files if they are not needed. Once you do a `git rebase -i master` this is good to merge.

vvolkl:
We should revisit the tests in the future though - there are a lot of duplicate test options file that don't cover everything.

vvolkl (approved):
Hi Jana,

thanks for this PR, this is really very good to have. I'm going to merge it as is, but a couple of comments to be adressed in the future maybe.

* As far as I can see, `FCCee_ECalBarrel_calibration.xml` differs from the main one really only by:
```
<           <rotation angle="InclinationAngle"/> <!-- inclination angle -->
< 	       <inner thickness="Pb_thickness" sensitive="true">
---
>           <rotation angle="InclinationAngle"/>  <!-- inclination angle -->
> 	       <inner thickness="Pb_thickness" sensitive="false">
99c100
< 	       <glue thickness="Glue_thickness" sensitive="true">
---
> 	       <glue thickness="Glue_thickness" sensitive="false">
102c103
< 	       <outer thickness="Steel_thickness" sensitive="true">
---
> 	       <outer thickness="Steel_thickness" sensitive="false">
106c107
<         <readout thickness="readout_thickness" sensitive="true">
---
```
I think we could add a flag in the detector constructor to avoid duplicating the rest of the xml.

* The docs mention the directories where the compact files are located, but could quickly explain the individual files and what they are for.

* In `FCCee_ECalBarrel.xml`, the comment still says `Common settings for the inclined EM calorimeter` but now it is really just the main xml file.

* Similarly the title of `FCCee_ECalBarrel_calibration.xml` is outdated

* Some displays of the detector would be good to have in the docs, I think. I'll try to create some.

PR #408, opened 2020-10-08
Remove outdated ACTS components in preparation of version upgrade

vvolkl:


gganis (approved):
Looks good to me.
Do I understand correctly that this removes completely, for the time being, the dependency on ACTS (acts or acts_core)?

vvolkl:
Hi Gerri, yes it completely removes it. But acts_core has changed so much since the version used here (v0.10, now it is the 1.x series and the project is renamed to just acts) that this will anyway have to be rewritten almost completely. And in the meantime, this allows us to install a recent version of acts.

PR #411, opened 2020-10-14
[evtgen] add additional arguments as used by stephane monteil

vvolkl:
See https://github.com/semkiv/fcc-generator/blob/6a0b9717a596c4dd9905855fd3941eb93b047804/src/generator/generator.cpp#L175


clementhelsens (commented):
Thanks, do you have evidence that this will fix the issue of not having specific decays?

vvolkl:
No, this is mostly for consistency with the previous work.

PR #412, opened 2020-10-23
bugfix in EvtGen decay

clementhelsens:
This is a bug fix to allow EvtGen to be called properly for the decay.
One thing to still sort out is the configuration file in the EvtGen constructor. It is currently stored on eos, and needs to be added to the Generation/data/


vvolkl (approved):
Good catch for this bug, Clement, Thanks. I'll update this to remove the absolute path for the dec file.

PR #419, opened 2020-11-04
Fix HEPMC units

vvolkl:
Explicitly use GeV in the HepMC GenEvent, instead of trying to find out and convert to the default units.


gganis (approved):
Looks good to me

PR #425, opened 2020-11-12
Topo clusteringto edm4hep branch oct21

BrieucF:

- port the calorimeter reconstruction 'topoClustering' algorithm to EDM4hep. Could only be compiled in an older version of the edm4hep branch as the head seems to be broken. 

To be done:
- port also the 'CellPositions' algorithm family which also rely on fcc::edm. The topo clustering should nevertheless run as such because the 'CellPositions' only use fcc::edm internally, they are used in the topoClustering to retrieve a dd4hep::position type.


vvolkl:
Excellent, thanks @BrieucF

PR #426, opened 2020-11-19
Add CaloHitCollection to store cluster cells

kjvbrt:
Hello all,

this pull request fixes crash, which occurs when one wants to attach cells to the calo clusters.
I'm not sure this is the right way to do it, found it by experimentation :)

### DIFF
--- a/Detector/doc/LArEcalInclined.md
+++ b/Detector/doc/LArEcalInclined.md
@@ -166,6 +166,7 @@ createClusters = CreateCaloClustersSlidingWindow("CreateClusters",
                                                  nEtaFinal = finE, nPhiFinal = finP,
                                                  energyThreshold = threshold)
 createClusters.clusters.Path = "CaloClusters"
+createClusters.clusterCells.Path = "CaloClusterCells"
 ~~~
 - Building of the towers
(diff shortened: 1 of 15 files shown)


vvolkl (changes requested):
Thanks for the patch! To make it easier to review, could you remove all the whitespace / formatting changes that don't touch change the actual code? You can reformat the code (best to run `clang-format`), but please do that in a separate commit or PR

kjvbrt:
All unnecessary formmating was removed.
I'm not sure about the failed test, the error looks unrelated to me.

BrieucF:
+1 

As discussed privately with @kjvbrt, some use case of this functionality do not need the attached cells to be stored permanently in the output rootfile (e.g. CorrectECalBarrelSliWinCluster). Transient storage can be investigated as a solution for that. Meanwhile, one can use `out.outputCommands = ["keep *", "drop CaloClusterCells"]` to save space in the output file if one do not need the attached cell for analysis

vvolkl:
LGTM.  Only just now fixed the tests, will merge now and check them on the master branch.

vvolkl (approved):

PR #429, opened 2020-12-01
Fix system id for caloCellPosition [temporary, should refactor this]

BrieucF:
Fix the system id in order to have meaningful positions for the ECAL barrel. This is a temporary fix, we should probably refactor this to avoid such a hard coding, possibly with another way to retrieve the system ID on the flight if it is possible. Or by having only one position tool [here](https://github.com/HEP-FCC/FCCSW/blob/master/Reconstruction/RecFCChhCalorimeter/src/components/CreateCaloCellPositions.h#L48) and by calling this algorithm once per sub-system in the config file.

### DIFF
--- a/Reconstruction/RecFCCeeCalorimeter/CMakeLists.txt
+++ b/Reconstruction/RecFCCeeCalorimeter/CMakeLists.txt
@@ -5,6 +5,16 @@ gaudi_subdir(RecFCCeeCalorimeter v1r0)
 
 gaudi_depends_on_subdirs(GaudiAlg FWCore Detector/DetInterface Detector/DetSegmentation Detector/DetCommon Reconstruction/RecInterface Reconstruction/RecCalorimeter)
 
+find_package(FastJet)
+find_package(ROOT COMPONENTS Physics Tree)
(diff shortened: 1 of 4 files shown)


faltovaj:
Hi Brieuc,
at the moment the algorithm is in the directory RecFCChhCalorimeter and for FCChh reference detector all IDs are set correctly. That's why I would not change the values in the code in here. A temporary solution could be to make a copy of the algorithm with correct IDs for FCCee in RecFCCeeCalorimeter. We can make "Issue" to keep in mind this is something to be fixed. What do you think?
Cheers,
Jana

BrieucF:
Hi Jana, good point, I will change that and create an issue!

BrieucF:
Updated, see issue #430

faltovaj:
Hi Brieuc,
thanks for the new version. I'm happy with that.
Cheers,
Jana

vvolkl:
LGTM. Could you also add those changes to https://github.com/HEP-FCC/k4RecCalorimeter?

vvolkl:
I recently updated the tests, will merge now and check the test results of the master branch.

PR #434, opened 2021-02-24
Allow the user to manually set the seed (useful for parallel production)

BrieucF:
This PR adds the possibility to manually set the seed used in `RndmGenSvc`. It can be used to avoid jobs starting with the same seed when launching a parallel production.


vvolkl (approved):
Looks good to me, thanks Brieuc. I'll add it to k4SimGeant4, where this code will migrate to.

PR #443, opened 2021-06-04
Add a full sim calo config running in key4hep

BrieucF:
This config works up to the cell with position, still need to adapt the clustering algorithms


vvolkl (commented):
Thanks, it's great to have these job configurations in hep-fcc repostories for testing. I've been rewriting these files in order for them to be actually modular, i.e. that you can do `fccrun particle_gun.py full_sim.py` and `fccrun pythia.py full_sim.py` and so on. I'll update the format of this one as well and give it a test.

vvolkl:
Superseded by #444

Issue #462, opened 2023-09-06
Adding Examples/scripts to Key4hep software stack


armin-ilg:
Hi

I want to use the FCC style file ([plotstyle.py](https://github.com/HEP-FCC/FCCSW/blob/master/Examples/scripts/plotstyle.py)) in another repository and would like to be able to import it like
```
sys.path.append($FCCSW + "/Examples/scripts")    # FCC Style from https://github.com/HEP-FCC/FCCSW/blob/master/Examples/scripts/plotstyle.py
from plotstyle import FCCStyle
```
or similarly, but in the Key4hep software stack only `FCCSW/Examples/options` is available. 
Could this be changed so that also the scripts folder is accessible? Is this here the correct place to ask this?

Thank you,
Armin

PR #463, opened 2023-09-28
Installing also Examples/scripts

kjvbrt:
Adding also `Examples/scripts` into FCCSW's share directory as suggested by @armin-ilg in https://github.com/HEP-FCC/FCCSW/issues/462.


armin-ilg:
Thanks, looks good to me as far as I can tell :-)

armin-ilg (approved):

PR #491, opened 2026-03-07
Fix tests.

scott-snyder:
Update tests to use geometry files from k4geo, rather than the obsolete ones in FCCDetectors.  The old ones fail due to errors in the type_flags elements.  These were previously not failing because dd4hep was hiding the errors.  However, a side effect of this change

https://github.com/AIDASoft/DD4hep/pull/1578

is that dd4hep is properly reporting the errors.

Also, run the ubuntu build only with the nightly on ubuntu24.

### DIFF
--- a/.github/workflows/key4hep-build.yaml
+++ b/.github/workflows/key4hep-build.yaml
@@ -15,7 +15,10 @@ jobs:
     strategy:
       matrix:
         build_type: ["release", "nightly"]
-        image: ["alma9", "ubuntu22"]
+        image: ["alma9"]
+        include:
+          - build_type: nightly
+            image: ubuntu24
       fail-fast: false
     runs-on: ubuntu-latest
     steps:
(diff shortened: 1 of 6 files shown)


tmadlener:
Since I just ran into similar failures in a different context. Can you tell me how this fixes issues with the type flags? This seems to only change detector models, but it doesn't touch any xml file. What am I missing here?

scott-snyder:
The DetType_ symbols are defined in the file detector_types.xml from DD4hep.  This should be included from the  top-level detector xml files, but this was not the case for most of the configurations in FCCDetectors.  The versions in k4geo do however do this.  Since the maintained versions are now the ones in k4geo, it seemed better to switch to using those versions rather than try to fix the ones in FCCDetectors.

tmadlener:
Ah, alright. That makes sense. Thanks a lot for the explanation.

PR #495, opened 2026-05-13
Remove PodioDataSvc, FCCDataSvc and PodioOutput

jmcarcell:
Following https://github.com/key4hep/k4FWCore/pull/392


jmcarcell:
CI on Alma 9 is green which is the only one that worked before (except for Ubuntu 22 release, which is a release from early 2025 I think.)

